PR #47, opened 2026-05-03
fix(traffic): mock banner warning + HERE source type

valerielinc-ops:
## Summary
- Changes mock data Callout from accent (blue) to warning (yellow) so users clearly see when they are looking at simulated estimates instead of live data
- Adds 'here' to the TrafficData source union type and hasLiveTrafficData() so the new HERE Maps provider is recognized as live data

## Test plan
- [ ] Navigate to /guida-frontaliere/tempi-attesa-dogana/ with no Firestore data → should see yellow warning banner
- [ ] With fresh Firestore data → should see green success banner


valerielinc-ops:
Duplicate of #45 which was already merged

PR #559, opened 2026-05-25
fix(traffic): filter clear-webcam wait outliers

valerielinc-ops:
## Summary
- add a webcam sanity filter for border wait outliers when HERE reports a red queue but the primary webcam is clear and the approach segment is free
- keep the existing positive webcam override that raises missed visible queues
- cover clear-webcam, poor-visibility, and approach-congested cases in traffic scheduler tests

## Verification
- npx vitest run tests/trafficScheduler.here-routing.test.ts tests/traffic-provider-migration.test.ts tests/trafficScheduler.test.ts
- node scripts/analyze-webcam-frame.mjs 02.0N
- npm run build


valerielinc-ops:
Closing in favor of a clean branch rebased from current main: codex/fix-border-wait-outliers-main.

PR #584, opened 2026-05-26
fix: prevent Pages retry cancellation

valerielinc-ops:
## Summary
- remove the second actions/deploy-pages retry from the Pages deploy job
- keep a single deploy-pages attempt so timeouts remain visible failures without canceling an in-progress Pages file sync
- remove the stale deployment_first environment URL fallback

## Root cause
Run 26427711236 uploaded a 9.6G tar / 1,722,656,147 byte Pages artifact, then deploy-pages timed out after its hard 600000 ms cap while Pages was still syncing. The workflow retry invoked deploy-pages again and the action canceled the active deployment at finished_file_sync. This PR removes that destructive retry path.

## Verification
- ruby YAML.load_file on .github/workflows/deploy.yml: yaml ok
- git diff --check
- actionlint .github/workflows/deploy.yml: only existing SC2005 on line 153 remains; no deployment_first/reference errors
- GitNexus detect changes: risk 0.00, no test gaps

Not merged while a Pages deploy is in progress, per deploy safety rule.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -117,7 +117,6 @@ jobs:
     timeout-minutes: 360
     environment:
       name: github-pages
-      url: ${{ steps.deployment.outputs.page_url || steps.deployment_first.outputs.page_url }}
     steps:
       # Free ~25-30 GB on the ubuntu-latest runner before any cache restore
       # or build output lands on disk. Without this the runner ships with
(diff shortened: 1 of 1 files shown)


valerielinc-ops:
Closing: deploy concurrency group already serializes pages-deploy; retry was causing collateral cancellation on slow syncs but auto-recovery loss not worth the change right now.

PR #2117, opened 2026-06-15
feat(translate): in-process Argos Translate mop-up tier ($0, unlimited, no HTTP race)

valerielinc-ops:
## Implementato

- **`scripts/local-mt-translate.py`** — in-process Argos Translate worker (JSONL stdin/stdout). `--install` installs the 6 `xx<->en` it/en/de/fr packages (Argos pivots the cross pairs — `de->it`, `it->fr`, … — through English, so no direct package needed); streaming path reads `{id,text,from,to}` per line and emits `{id,text}` / `{id,error}`. A network-free `models_ready()` check is used on the translate path so it never re-hits the HuggingFace index when models are already installed (the network fetch hangs on a constrained runner).
- **`scripts/local-mt-mopup.mjs`** — Node orchestrator that scans `data/jobs/by-crawler/*.json`, finds jobs still needing translation by **importing the exact predicate from `relocalize-pending-jobs.mjs`** (`isIncomplete` + `reconcileRetranslationState`, plus the same `needsRetranslation`/suppression gate), computes only the genuinely-missing locale `title`/`description` slots, runs the Python worker **once** over the whole batch (models loaded once), and merges results back into the same slices with the same write guards relocalize uses (never overwrite a good translation, never write a source-copy, never write empty). Clears `needsRetranslation` via `reconcileRetranslationState` **only when the locale actually became complete**. Wall-clock budget `LOCAL_MT_TIME_BUDGET_MS` (default 280min) stops new writes before the 350min job timeout. **Does NOT touch `scripts/lib/free-translate.mjs`** (parallel-agent file) — fully decoupled, shells to Python.
- **`.github/workflows/translate-pending.yml`** — two new steps after “Scatter changes back to per-crawler slices” and before the existing “Commit translations”: an Argos install step (`pip install argostranslate` + `--install`) and the mop-up step (`node scripts/local-mt-mopup.mjs --max-jobs …`). Both `continue-on-error: true` and gated by `vars.LOCAL_MT_DISABLE != '1'` → strictly additive; the existing commit captures the slice output.
- **`.gitignore`** — add `__pycache__/` (Python bytecode artifact).

**Engine choice — Argos Translate (not the LibreTranslate Docker service, not raw CTranslate2):** Argos *is* CTranslate2/OPUS-MT int8 under the hood (same models LibreTranslate wraps) but loaded as a **direct Python library** — no HTTP server, no `/languages` healthcheck. That healthcheck race is exactly what made the old self-hosted LibreTranslate service 100% dead on the runner (1483 timeouts, 0 successful translations): the healthcheck passed before models loaded, the first `/translate` cold-load blew the 30s client timeout, and the cascade abandoned it (death spiral). In-process loading removes the HTTP layer and the race entirely. Raw CTranslate2+transformers would be marginally faster but needs manual model download/convert + tokenizer wiring — worse reliability-per-effort for CI. NLLB-200/madlad are bigger/slower on a 2-core CPU. Argos wins on reliability-per-effort.

## Non implementato (ancora)

- **CI behaviour is unvalidatable pre-merge.** The model download (~a few hundred MB from the HuggingFace CDN), the `pip install` of `argostranslate`+`ctranslate2`+`sentencepiece`, and the actual mop-up wall-time only run post-merge on the `translate-pending` workflow. **Revert-trigger:** if the next live run OOMs, times out, the Argos install step fails CDN-download, or the mop-up produces **0 net new translations** → revert this PR (or set repo var `LOCAL_MT_DISABLE=1` to disable the two steps without a code change) and iterate on model-cache/runner sizing. The steps are `continue-on-error`, so a failure there cannot fail the workflow — worst case they are a no-op.
- **Model-download size/time risk:** ~6 packages downloaded once per run (no cross-run cache wired yet). If this proves slow, a follow-up can add `actions/cache` keyed on the Argos package dir (`~/.local/share/argos-translate`). Deferred (out of scope for the first landing; correctness first).
- **Quality is mop-up grade, by design.** Argos is lower-quality than DeepL/Azure (e.g. German abbreviations like “Stv.” not expanded). This is intentional — it only runs on leftovers the premium/public cascade couldn’t finish, and the downstream `mark-locale-mismatched-jobs.mjs` / `isIncomplete` detector re-flags any genuinely-bad output on the next run, so it can never regress a good translation.
- **No new shared-module extraction / no sibling class-fix.** This adds a *new tier*; it is not a regex/guard/threshold fix of an existing construct. `node scripts/ci/check-sibling-patterns.mjs` surfaced only generic shared identifiers (`MAX_JOBS`, `BY_CRAWLER_DIR`, `MIN_DESC_CHARS`, `normalizeCompanyKey`, `exitCode`) — common constants across many scripts, not a shared antipattern — so there is nothing to sweep.

## Test plan

Verified the engine and the full orchestrator locally (in a Python venv with `argostranslate`):

- [x] `python3 -c 'import argostranslate'` + `--install` → `📦 Argos model install complete: 6/6 packages ready`, all 4 locales installed.
- [x] **Real translations (not echoes)** through the JSONL protocol, warm batch `4 ok, 0 failed in 5.4s`:
  - **DE→IT (description):** `Wir suchen einen erfahrenen Softwareentwickler für unser Team in Zürich. …` → `Stiamo cercando uno sviluppatore di software esperto per il nostro team a Zurigo. …`
  - **IT→DE (title):** `Consulente Assicurativo` → `Versicherungsberater`
  - **IT→FR:** `Responsabile delle vendite per la regione Ticino` → `Directeur des ventes pour la région du Tessin` (pivoted it→en→fr)
  - **FR→DE:** `Nous recherchons un comptable expérimenté pour notre bureau de Genève.` → `Wir suchen einen erfahrenen Buchhalter für unser Büro in Genf.`
- [x] **End-to-end orchestrator** (`node scripts/local-mt-mopup.mjs --max-jobs 3`): scanned slices, queued 5 field translations, `5 ok / 0 failed`, wrote 5 locale fields to 1 slice, cleared 1 `needsRetranslation` flag. Inspected the diff (real translations e.g. `"en": "Head of Restoration"`, `"fr": "Chef de la restauration"`), then reverted the test data.
- [x] `node --check scripts/local-mt-mopup.mjs` ✓ · `python3 -m py_compile scripts/local-mt-translate.py` ✓ · workflow YAML parses ✓.
- [x] `npx vitest run tests/relocalize-is-incomplete.test.ts tests/relocalize-suppression-loop.test.ts` → 11 passed (confirms importing relocalize’s exports doesn’t trigger its `main()` or break its contract).
- [ ] Live `translate-pending` run actually fills leftover locale fields and commits via the existing slice-commit path (unvalidatable pre-merge — see revert-trigger above; observe on `main` post-merge).


valerielinc-ops:
⚠️ **collision-risk**: questa PR tocca file funnel-critical condivisi con la PR #2128: `build-plugins/jobsSeoPagesPlugin.ts`. La seconda a raggiungere il merge DEVE prima rebasare oltre l'altra (`git merge origin/main` dopo che l'altra è mergiata) — l'auto-merge è bloccato finché `collision-risk` + dietro main. _Segnale deterministico da pr-collision-detector.yml (zero-Claude)._

valerielinc-ops:
⚠️ **collision-risk**: questa PR tocca file funnel-critical condivisi con la PR #2127: `build-plugins/jobsSeoPagesPlugin.ts`. La seconda a raggiungere il merge DEVE prima rebasare oltre l'altra (`git merge origin/main` dopo che l'altra è mergiata) — l'auto-merge è bloccato finché `collision-risk` + dietro main. _Segnale deterministico da pr-collision-detector.yml (zero-Claude)._

valerielinc-ops:
⚠️ **collision-risk**: questa PR tocca file funnel-critical condivisi con la PR #2057: `build-plugins/jobsSeoPagesPlugin.ts`. La seconda a raggiungere il merge DEVE prima rebasare oltre l'altra (`git merge origin/main` dopo che l'altra è mergiata) — l'auto-merge è bloccato finché `collision-risk` + dietro main. _Segnale deterministico da pr-collision-detector.yml (zero-Claude)._

Issue #2159, opened 2026-06-15
SearchAtlas audit: remaining actionable SEO fixes (dup meta-desc, dup titles, hreflang self-ref)


valerielinc-ops:
Tracking the actionable remainder of the SearchAtlas site audit (id 141162) after PR #2158 (which fixed `invalid_rel_found`, 6,111 pages). Root causes pinned below; each is a focused PR. The rest of the audit is deliberate-by-design or false-positive (documented in #2158 `## Non implementato`).

### 1. Duplicate / missing meta descriptions — 1,547 + 51 pages (highest real-SEO value)
- Root cause: `SEO_SECTION_DESCRIPTION_KEY_MAP` in `services/seoService.ts:956` is incomplete. ~40 SPA sections (whatif, ral, bonus, parental-leave, jobs, nursery, glossario, faq, …) have a title key but **no** description key, so they fall through to `buildLocalizedSeoFallbackDescription()` (`:1015`) → the same generic sentence per locale across all of them.
- These sections are **indexable** (`<meta name="robots" content="index,follow">` confirmed live; raw-HTML canonical is the homepage but `seoService.updateMetaTags` rewrites it client-side, and the audit renders JS).
- Fix: add unique `seo.<section>.description` keys (×4 locales it/en/de/fr) and wire them into the map. Substantial localized-copy task — do it with real per-section copy, not near-dup templates (templated dupes won't clear the flag).

### 2. Duplicate `<title>` — 231 pages
- Suspected: employer-hub title built with `count: 0` hardcoded — `build-plugins/jobsSeoPagesPlugin.ts:3290` (`ctrTitle` → `buildEmployerHubTitle({ count: 0 })`).
- Caveat: titles already vary by company name, so the dup set needs confirmation against the per-URL list before touching the builder. Verify first.

### 3. `no_self_ref_hreflang` — 612 pages
- Root cause: some feature plugins emit hreflang without routing through the canonical `buildHreflangEntries()` / `renderHreflangTags()` helper (`build-plugins/shared/hreflang.ts:110`), so the current locale's self-reference is missing.
- Fix: identify the offending plugins (orphan-query, fuel-daily, weekly-employers, border-wait, health-premiums are candidates) and route them through the shared helper.

### Explicitly NOT to do (gated / by design)
[...]

PR #2163, opened 2026-06-15
fix(seo): unique meta descriptions for SPA sections (SearchAtlas non_unique_meta_desc)

valerielinc-ops:
Part of #2159

## Implementato
- Root cause: `SEO_SECTION_DESCRIPTION_KEY_MAP` in `services/seoService.ts` covered only 35 of the 77 sections that have a title key in `SEO_SECTION_TITLE_KEY_MAP`. The 42 uncovered sections fell through to `buildLocalizedSeoFallbackDescription()`, which emits one generic sentence per locale across all of them → duplicate en/de/fr meta descriptions (SearchAtlas `non_unique_meta_desc`). IT was unaffected because IT pulls per-section copy directly from `seo-pages.ts`.
- Wired 42 new `seo.<section>.description` entries into `SEO_SECTION_DESCRIPTION_KEY_MAP` (no removals, no logic change). The two maps are now in full parity (77/77, 0 missing).
- Added 168 new locale keys (42 sections × 4 locales) to `services/locales/{it,en,de,fr}-core.ts`, each a genuinely unique, topical meta description (not templated near-dupes): withholding tax, Permit G/B, AVS/LPP, LAMal, 13th/14th salary, TFR, unemployment/PD U1, parental leave, tax returns IT/CH, ristorni, mortgage compare, border traffic/map, schools, municipalities, etc.
- All 168 descriptions verified unique within each locale and with 0 collisions against the 16 pre-existing `seo.*.description` keys; all sit in the 140–160 visible-character band.
- Sections fixed (42): whatif, ral, bonus, parental-leave, salaryQuiz, jobs, tax-return-italia, tax-return-svizzera, nursery, renovation, quiz, firstDay, border, unemployment, border-map, livingCH, livingIT, companies, schools, places, municipalities, calendar, morning, ristorni, trafficHistory, unemploymentStats, mortgageComparison, glossario, faq, dialetto, sitemap, contracts, tfr-calculator, permit-quiz, tredicesima, weekly-digest, tool-of-week, feedback, forum, dashboard, gamification, privacy.

## Non implementato (ancora)
- IT-locale section descriptions were NOT changed — IT already resolves unique per-section copy from `seo-pages.ts` via `metadata.description`, so it was never part of the duplicate-description flag. The new IT keys are added for locale parity/consistency only.
- Job-board / blog / landing chunk pages out of scope — they have their own description sources (`meta-descriptions.ts`, `seo-blog*.ts`, `seo-landing.ts`) and were not implicated in this flag.
[...]

PR #2164, opened 2026-06-15
fix(seo): self-referential hreflang on pages bypassing the shared helper (SearchAtlas)

valerielinc-ops:
## Implementato

- **Root cause pinned to the SPA runtime, not the build plugins.** SearchAtlas's `no_self_ref_hreflang` (612 pages, #2159 item 3) is produced by `updateHreflangTags(route)` in `services/seoService.ts`, which rebuilt every hreflang tag from `buildAllLocalePaths(route)` → `buildPath(route, locale)`. SearchAtlas renders JS, so it observed this clobbered runtime block instead of the correct server-rendered one.
- **Why `buildPath` produces the wrong self-href on these pages:** `staticOverlay` landing routes (recency/today landings, fuel-daily, border-wait, salary-stats, profession-canton, publisher ads, …) do NOT carry the specific landing slug in `AppRoute`, so `buildPath(route, locale)` collapses to the generic tab root. Verified empirically: `parsePath('/cerca-lavoro-ticino/ultimi-3-giorni/')` → `staticOverlay: true`, and `buildAllLocalePaths(route).it === '/cerca-lavoro-ticino/'` (not the landing URL). The self (it) hreflang therefore pointed at the generic root, never the page's own URL → "no self-ref".
- **Fix (surgical, 1 source edit):** `updateHreflangTags` now early-returns on `route.staticOverlay`, leaving the server-rendered hreflang block intact. That block is already correct, self-referential, 4-locale + x-default — emitted by the build plugins via the shared `renderHreflangTags()` / `buildHreflangEntries()` helper (`build-plugins/shared/hreflang.ts`). This mirrors the existing canonical / `og:url` logic in the same function, which already uses `window.location.pathname` for `staticOverlay` routes (lines ~4088-4090) precisely because the route can't reconstruct the slug.
- **Regression test:** `tests/seo-static-overlay-hreflang-selfref.test.ts` seeds the correct static hreflang block, navigates to a real recency landing URL, calls `updateMetaTags`, and asserts the `it` + `x-default` self-references still point at the landing URL. Confirmed it FAILS without the guard (self href clobbered to the generic root) and PASSES with it.
- **Est. pages covered:** all `staticOverlay` landing families that ship a server-rendered hreflang block (recency/today, fuel-daily, border-wait, salary-stats, profession-canton, salary-hub, publisher ads, …) — the SearchAtlas-flagged 612 are a subset of these.
- **Audit safety net untouched:** `hreflangPostprocessPlugin` / `postWalkCoordinatorPlugin` broken-alternate stripping is unchanged.

I verified the build-plugin emitters named as candidates in #2159 (orphan-query, fuel-daily, weekly-employers, border-wait, health-premiums) and the other hand-rolled emitters (sectionPages, careerLandings, costOfLiving, comparisons, faqHub, nursing, profession, jobRecency, jobSector, ogPages, publisherAd, sectionPages, seoHubs, weatherCity/Alert, borderMunicipality, jobsSeoPagesPlugin, relatedSearchClusters): every one either routes through the shared helper or maps over the full locale list so the self-locale href equals the canonical. None omits self-ref in the static HTML. The bug is exclusively the SPA runtime overwrite.

## Non implementato (ancora)

- Other #2159 items (duplicate/missing meta descriptions ~1,547+51, duplicate `<title>` ~231) — separate focused PRs per the issue, out of scope here.
- No change to the build-plugin `renderHreflangTags` call sites: they are already correct (verified), so touching them would be churn. The `check-sibling-patterns.mjs` surfacer lists `renderHreflangTags` / `AppRoute` token-sharing files, but those share the *solution* (the shared helper) or just the route type — none shares the actual bug class (SPA route-derived hreflang overwrite on staticOverlay). The fixed function is the only runtime hreflang rebuilder (`grep` confirms `buildAllLocalePaths` has a single non-test caller).
- Live confirmation that SearchAtlas drops the flag: deferred to the post-merge audit cycle (SearchAtlas re-crawl), since it renders JS and cannot be exercised pre-merge without a deploy.

## Test plan

- [x] `npm ci`
- [x] `npx vitest run tests/seo-static-overlay-hreflang-selfref.test.ts` — 2 passed; verified it fails without the guard and passes with it.
- [x] `npx vitest run tests/seo/hreflang.test.ts` — 49 passed.
- [x] `npx vitest run tests/seo-noindex-filter-variants.test.ts tests/seo-localization.test.ts tests/seo-html-lang-sync.test.ts tests/router.test.ts tests/app-lite-shell.test.tsx tests/soft-landing-seo.test.ts tests/seo/cathedral-hreflang-cross-locale.test.ts tests/seo/cathedral-hreflang-x-default.test.ts tests/seo/legacy-city-hub-hreflang.test.ts` — all green (557 passed in the combined run).
- [x] GitNexus impact (`updateHreflangTags`, upstream): LOW risk, single caller `updateMetaTags`.
- [ ] Post-deploy: SearchAtlas re-crawl shows `no_self_ref_hreflang` count drop on the staticOverlay landing families (verifiable only after deploy + re-audit).

`tests/hreflang-no-broken.test.ts` is a dist-only post-build test (skips without a built `dist/`); not runnable here since full vite build is forbidden (OOM). It runs in CI post-merge.

Part of #2159


valerielinc-ops:
⚠️ **collision-risk**: questa PR tocca file funnel-critical condivisi con la PR #2163: `services/seoService.ts`. La seconda a raggiungere il merge DEVE prima rebasare oltre l'altra (`git merge origin/main` dopo che l'altra è mergiata) — l'auto-merge è bloccato finché `collision-risk` + dietro main. _Segnale deterministico da pr-collision-detector.yml (zero-Claude)._

PR #2205, opened 2026-06-15
feat(translate-pending): never lose work on timeout OR any error (budget + always()-guarded mop-up/commit)

valerielinc-ops:
## Implementato

- **Opt-in wall-clock budget in the shared crawler's AI-localization loop** (`scripts/lib/shared-jobs-crawler.mjs`, in `main`/`runSharedCrawlerPipeline`, around the `runWithConcurrency(selectedQueue…)` block). Reads a NEW env var `JOBS_AI_LOCALIZATION_TIME_BUDGET_MS`:
  - **Default UNSET / 0 / non-finite ⇒ NO budget (unlimited).** `Number.isFinite(raw) && raw > 0 ? raw : 0`. The 300+ normal dedicated crawlers never set it → completely unaffected (zero behavior change). This opt-in default is the critical safety property.
  - Loop start captured (`localizationLoopStart = Date.now()`) just before the queue runs. The per-item callback guards at its top: when `budget > 0 && (Date.now() - start) > budget`, it STOPS starting new jobs by early-returning the job UNCHANGED (`return { fp: fingerprintJob(job), enriched: job }`), which maps the job to itself in `enrichedMap` (a no-op) so it keeps its `needsRetranslation`/incomplete flag and the next run retries it. No AI call is made for deferred jobs. In-flight items finish normally.
  - Logs once: `⏰ [localize] time budget <N>min reached — N jobs deferred to next run` (count = `selectedQueue.length - index` at first deferral; `runWithConcurrency` pulls items in order so the unstarted tail is exactly the deferred set).
  - Jobs localized BEFORE the budget hit keep their results — they flow through `enrichedMap` → `merged` as today, and the existing per-company incremental `fs.writeFileSync` in `relocalize-pending-jobs.mjs` (`syncTranslationsToCrawlerFile` after each `runSharedCrawler([key], …)`) captures them into the per-crawler slices that get committed. Verified.
- **Workflow env wired** (`.github/workflows/translate-pending.yml`, "Phase 2: Translate pending jobs" step): `JOBS_AI_LOCALIZATION_TIME_BUDGET_MS: '16800000'` (280 × 60 × 1000), matching the Argos mop-up's `LOCAL_MT_TIME_BUDGET_MS` default and well under `timeout-minutes: 350` — leaving headroom for assemble-after-stats, scatter, mop-up and the commit steps.
- **always()-guard hardening** (`.github/workflows/translate-pending.yml`): the "Scatter changes back to per-crawler slices" step was guarded only by `if: inputs.skip_translate != true` → it would be SKIPPED if Phase 2 errored. Changed to `if: always() && steps.checkout.outcome == 'success' && inputs.skip_translate != true` so the final scatter still runs after a partial/cut run. "Log translation stats (after)" and "Commit translations" were ALREADY `always() && steps.checkout.outcome == 'success' && …`-guarded — verified, no change needed. The commit uses the existing `scripts/lib/git-commit-data.sh --slice-only` over `data/jobs/by-crawler/`, i.e. the incremental per-crawler writes.

## Non implementato (ancora)

- **Effect not validatable pre-merge.** Only a live, timeout-length translate-pending run (a backlog big enough for ONE company's in-process `runSharedCrawler` call to approach 350min) proves no-loss. The `pull_request` runner cannot reproduce that wall-clock. **Revert-trigger:** if a normal dedicated crawler somehow regresses (defers/skips localization despite never setting `JOBS_AI_LOCALIZATION_TIME_BUDGET_MS`) → revert; the default-off guard (`raw > 0 ? raw : 0`) is designed to make this impossible by construction.
- **The Argos mop-up (`scripts/local-mt-mopup.mjs`) already had its own budget** (`LOCAL_MT_TIME_BUDGET_MS`, default 280min, commits partial) — this PR closes the same gap for the cascade step (the shared crawler's AI loop), which had NO internal time guard. The `check-sibling-patterns` surfaced it as a `LOCAL_MT_TIME_BUDGET_MS` token sibling; intentionally not modified — it is the already-correct reference, not a sibling needing the same fix.
- **`fingerprintJob` siblings** (`scripts/update-axa/eoc/lafonte/swisscom-jobs.mjs`, `dedicated-crawler-common.mjs`, `backfill-slug-aliases.mjs`, `dry-run-target-cantons-flip.mjs`) surfaced by `check-sibling-patterns` only USE `fingerprintJob` (a token my diff also references in the unchanged early-return). No shared antipattern — they have no localization-loop wall-clock concern. Not touched.

## Test plan

- [x] `node --check` on both edited files — OK.
- [x] Dynamic import smoke of `shared-jobs-crawler.mjs` — `runSharedCrawlerPipeline` resolves as `function`.
- [x] `npx vitest run` on localization/relocalize tests (`relocalize-suppression-loop`, `relocalize-is-incomplete`, `job-localization-pipeline`, `dedicated-crawler-localization-pipeline`, `dedicated-crawlers-localization-coverage`) — **18 passed**.
- [x] YAML validity (`python3 -c "import yaml; yaml.safe_load(...)"`) — OK.
- [x] `node scripts/ci/check-sibling-patterns.mjs` after `git add -A` — candidates reviewed (see Non implementato); no same-class fix omitted.
- [ ] Live timeout-length translate-pending run confirms no work lost (only verifiable post-merge on a large backlog; see Non implementato revert-trigger).

### DIFF
--- a/.github/workflows/translate-pending.yml
+++ b/.github/workflows/translate-pending.yml
@@ -153,14 +153,12 @@ jobs:
           # Self-hosted LibreTranslate removed (see note above the steps): it was 100%
           # dead on the runner. The cascade now uses premium tiers + MyMemory + PUBLIC
           # LibreTranslate, which were already doing all the real work.
-          # Wall-clock budget for the in-process AI-localization loop inside the shared
-          # crawler. 280min — matching the Argos mop-up (LOCAL_MT_TIME_BUDGET_MS) and
-          # well under this job's timeout-minutes:350 — leaves headroom for assemble,
-          # scatter, mop-up and the always()-guarded commit steps to run. Once exceeded,
-          # the crawler stops starting NEW jobs (leaving them flagged for the next run)
-          # so the step returns NORMALLY instead of being killed at the 350min timeout
-          # (a killed job does not reliably run the later commit step → lost work).
-          JOBS_AI_LOCALIZATION_TIME_BUDGET_MS: '16800000'  # 280 * 60 * 1000
+          # NB: the AI-localization wall-clock budget is set DYNAMICALLY per-company by
+          # relocalize-pending-jobs.mjs (JOBS_AI_LOCALIZATION_TIME_BUDGET_MS = remaining
+          # time until its 250min run-wide cascade deadline), so it is intentionally NOT
+          # hard-coded here — a static value reset the budget per company and could blow
+          # past the 350min job timeout (review #2205 🔴). The cascade stops by ~250min,
+          # leaving ~100min for the mop-up + always()-guarded commit/scatter/slug/deploy.
         run: node scripts/relocalize-pending-jobs.mjs --max-jobs ${{ inputs.max_jobs || '100' }} ${{ inputs.company_key && format('--company-key {0}', inputs.company_key) || '' }}
 
       - name: Log translation stats (after)
@@ -211,6 +209,12 @@ jobs:
       - name: "Phase 2 mop-up: local MT (Argos Translate, in-process)"
         if: always() && steps.checkout.outcome == 'success' && inputs.skip_translate != true && vars.LOCAL_MT_DISABLE != '1'
         continue-on-error: true
+        env:
+          # Bound the mop-up to the window left after the cascade's ~250min deadline
+          # (80min) so it self-stops + commits partial well before the 350min job
+          # timeout — instead of being externally killed mid-batch and losing its
+          # work (observed in validation run 27544487773). Default would be 280min.
+          LOCAL_MT_TIME_BUDGET_MS: '4800000'  # 80 * 60 * 1000
         run: node scripts/local-mt-mopup.mjs --max-jobs ${{ inputs.max_jobs || '100' }}
 
       - name: Commit translations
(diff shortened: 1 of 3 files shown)


valerielinc-ops:
### Esteso: "qualsiasi errore", non solo timeout

Su richiesta: il lavoro fatto fino a quel punto non deve perdersi su **nessun** errore (non solo timeout). Aggiunto in questo commit:

- **Mop-up Argos + Install Argos** ora `always() && checkout.outcome==success`-guarded → il mop-up gira e committa i leftover **anche se lo step cascade "Phase 2: Translate" va in errore** o viene tagliato dal budget (prima era skippato su errore cascade). Resta `continue-on-error` (non può far fallire il workflow).

Catena di resilienza completa (no-loss su timeout **e** errore):
1. `relocalize-pending-jobs.mjs` processa **per-company** con `try/catch` per company e **scrive su disco dopo ogni company** (`DATA_JOBS` + slice) → ogni company è un checkpoint; l'errore di una non perde le altre già scritte.
2. Budget wall-clock (questa PR) → il cascade **ritorna** prima del timeout 350min invece di essere ucciso.
3. `Log stats after` + `Scatter` + `Commit translations` sono `always()`-guarded → qualsiasi fallimento di step committa comunque ciò che è arrivato su disco.
4. Mop-up Argos `always()` + `continue-on-error` → ultima rete di salvataggio anche su errore-cascade.

YAML validato. Effetto pieno verificabile solo su run live (revert-trigger se un crawler normale regredisce: il budget è opt-in default-off).

PR #2639, opened 2026-06-22
perf(build): locale-gate jobSectorPagesPlugin so matrix shards skip non-owned-locale render

valerielinc-ops:
## Implementato

Adds a per-locale render-skip gate to `build-plugins/jobSectorPagesPlugin.ts`: `if (!shouldEmitLocale(locale)) continue;` at the top of the inner `for (const locale of LOCALES)` loop, plus the `shouldEmitLocale` import. This is the same render-skip `jobsSeoPagesPlugin` already uses (39 refs) and the documented follow-up in `shared/localeEmitFilter.ts` ("Gating those direct writers for additional CPU savings is a documented follow-up").

**Why (measured, not assumed).** Per-plugin `closeBundle` timing pulled from the last successful deploy (run `27929801749`) shows `job-sector-pages` costs **~150s on BOTH the `it` and `en` shards** (IT 149.8s / EN 154.4s). Near-equal time on two different shards = the plugin re-renders **all four locales on every shard**, then `scripts/ci/prune-locale-shard.mjs` deletes 3/4 of the output. The WriteCollector/prune only discard the *writes* — the expensive `buildSectorLandingHtml` render already ran. This is ~4× redundant compute.

**Zero page loss / byte-identical owned output** (the hard constraint):
- **Default all-locale build** (`BUILD_LOCALE` unset): `EMIT_ALL_LOCALES` → `shouldEmitLocale` always returns `true` → the gate is a **literal no-op**. Dev/vitest output unchanged (the gate cannot affect any test that runs without `BUILD_LOCALE`).
- **`it`/main shard** (`BUILD_LOCALE=it`): the `it` iteration always runs, so the `sitemap-sector.xml` entry — built **only** in the `locale === 'it'` branch, with its 4-locale `xhtml:link` alternates from `buildSectorHubPath` path-builders (independent of whether en/de/fr iterations ran) — stays complete. The en/de/fr renders are skipped but were pruned by `prune-locale-shard.mjs` anyway → final `it`+shared output is byte-identical.
- **en/de/fr shards**: only the owned locale renders; the (pruned-anyway) root `sitemap-sector.xml` is not built, matching `push-locale-shard` ownership.

**Saving:** ~112s off the `it` shard (the matrix wall-clock gate / time-to-live) and ~450 compute-s per deploy.

## Non implementato (ancora)

This is the surgical, verified-safe first slice of a larger, measured opportunity. The remaining items are **out of scope for this PR** and tracked as follow-ups:

- **Sibling ungated per-locale-emit plugins** (same antipattern class, AGENTS.md #6): `fuelDailyPagesPlugin` (~14s), `weeklyEmployersPlugin` (~11s), `jobMarketSnapshotPlugin` (~11s), `healthPremiumsLandingPlugin` (~3s), `borderWaitPagesPlugin` (~23s), `seoHubsPlugin`, `staticPagesPlugin` (~27s). **Deferred, not ignored:** unlike `job-sector` (whose sitemap is built solely in the `locale === 'it'` branch — verified safe), each of these builds its sitemap/shared-state with a *different* per-loop structure (e.g. some `push` sitemap entries inside every iteration, some accumulate cross-locale state). Each needs individual sitemap-accumulation verification before gating, using **this PR as the proven template**. Total residual ~90s — lower priority than `job-sector`'s 150s. Gating them prematurely in one pass risks silently dropping sitemap/hreflang entries (the exact SEO regression this PR is careful to avoid).
- The flagged siblings that **already** gate correctly (`jobsSeoPagesPlugin`, `hreflangPostprocessPlugin`, `jobOgImagesPlugin`, `orphanQueryLandingPlugin`, `relatedSearchClustersPlugin`, `salaryHubIndexLinkPlugin`, `professionLandings*`, `hreflangGuard`, `batchWrite`) are **not** antipattern — no change needed.
- **Bigger levers identified in the same study, separate concerns → separate PRs:**
  - `related-search-clusters` recomputes its locale-independent gram-index/postings prep (~200s) from scratch on every shard (`RELATED_SEARCH_CLUSTERS_NO_CACHE=1`, a debug flag, forced in prod; build-locale also doesn't cache `.cache/related-search-clusters`). Fix = hoist that prep into the `prep` job and ship it in `prepared-snapshot.tgz` (compute once). Biggest single IT-path win, but touches build orchestration + snapshot + OOM-sensitive prepass → its own PR.
  - `blog-image-cdn-finalize` (~200s) walks the **full** dist (all 4 locales) on every shard before the prune. Fix = scope the walk to owned-locale subtrees / reorder after prune.
  - `validate-dist` (57min, gates publish/indexing): rehydrates en/de/fr via `git clone --depth 1` (~14min) instead of downloading the same-run shard artifacts; its 4 data-only source-validators (~14min) run serially after a rehydrate they don't need. Fixes = same-run `download-artifact` (preserve the #1607 dir+homepage guard) + overlap/relocate the data-only validators.
  - Correctness (not speed): `Push locale shard` + offload run under `continue-on-error: true` → a failed shard push ships a stale locale while deploy still goes green.

## Test plan

- [ ] vitest CI green (the gate is a no-op without `BUILD_LOCALE`, so `tests/seo/job-sector-page-weight.test.ts` + `tests/seo/cathedral-sector-hubs.test.ts` exercise identical output).
- [ ] Post-merge: next matrix deploy's `validate-dist` / `validate-live` confirm critical sector pages + `sitemap-sector.xml` + hreflang are intact on the live `it`+shared output.
- [ ] Post-merge: confirm the `it` shard's `Build` step time drops (~150s → ~38s for this plugin) in the build-locale (it) job log.


valerielinc-ops:
⚠️ **collision-risk**: questa PR tocca file funnel-critical condivisi con la PR #2632: `build-plugins/jobSectorPagesPlugin.ts`. La seconda a raggiungere il merge DEVE prima rebasare oltre l'altra (`git merge origin/main` dopo che l'altra è mergiata) — l'auto-merge è bloccato finché `collision-risk` + dietro main. _Segnale deterministico da pr-collision-detector.yml (zero-Claude)._

PR #3188, opened 2026-07-01
fix(job-canon): CDN-offload canton-drift map + restore dead IT 404 recovery

valerielinc-ops:
## Implementato

**A. CDN-offload `dist/job-canon/*.json`** (canton-drift 404-recovery shard map, ~29 MB / 400 shards), same pattern already used for `/data` and `/og`:
- `scripts/lib/deploy-it-pages-prep.sh` (`step_push_cdn`): stage `dist/job-canon` into the CDN push payload; updated the "no offloadable assets" guard to include it.
- `scripts/offload-generated-images-cdn.mjs`: new delete-only phase (no static HTML ref exists to rewrite — confirmed via full-repo grep) that removes `dist/job-canon` once the CDN push has staged it.
- `infra/cloudflare-worker/locale-router.js`: `recoverCantonDriftOrphan` now fetches shards from `https://cdn.frontaliereticino.ch/job-canon/<sk>.json` instead of same-origin (`JOB_CANON_CDN_BASE` const, mirrors the existing duplication pattern of `JOB_DETAIL_RE`).
- `public/404.html`: same fetch now prefixed with a CDN-absolute base (`JOB_CANON_CDN`).
- `build-plugins/jobCanonRedirectMapPlugin.ts` + `vite.config.ts`: updated stale header comments that described job-canon as deliberately kept same-origin/root-emitted (that rationale is now obsolete/inverted).

**B. Fixed separately-discovered production bug**: `step_spa_fallback` in `deploy-it-pages-prep.sh` unconditionally did `cp dist/index.html dist/404.html` on every IT-locale deploy, silently overwriting the richer `public/404.html` (canton-drift recovery + generic SPA-redirect-restore) with a plain SPA-fallback copy — verified live that `frontaliereticino.ch/404.html` served the homepage instead of the recovery script. Fixed with an existence guard: `[ -f dist/404.html ] || cp dist/index.html dist/404.html`. Confirmed via `.github/workflows/deploy.yml` step ordering that `push-ticino-shard.sh` (which stages the Ticino/`it` shard's own `404.html`) runs *before* this step, so only the main IT apex was ever affected; en/de/fr shards don't ship a static `404.html` at all (they rely on the Worker's real 301) and are unaffected.

**Testing**: ran the 5 directly-affected suites (`job-canon-redirect-map`, all 4 `locale-router-*-301` tests) plus the 2 suites that directly invoke `offload-generated-images-cdn.mjs` (`cdn-data-preconnect-inject`, `job-detail-seed`) — 62/62 green. Syntax-checked `deploy-it-pages-prep.sh` (`bash -n`), `offload-generated-images-cdn.mjs` + `locale-router.js` (`node --check`), and the 3 inline `<script>` blocks in `public/404.html` (`new Function`).

**GitNexus**: `impact` on `recoverCantonDriftOrphan` → LOW risk, 0 tracked callers. `step_spa_fallback`/`step_push_cdn` (bash functions) are outside GitNexus's JS/TS index — blast radius for those instead verified manually via the `deploy.yml` step-ordering trace above. `detect_changes` returns 0 changed/affected against the indexed main tree (worktree-local diff, expected — not a gap in this PR's coverage).

## Non implementato (ancora)

Nessuno. Both original asks (CDN-feasibility investigation → implement, and the discovered dead-404-recovery bug) are fully implemented and tested in this PR.

### DIFF
--- a/build-plugins/jobCanonRedirectMapPlugin.ts
+++ b/build-plugins/jobCanonRedirectMapPlugin.ts
@@ -21,9 +21,11 @@
  * flat `{slug: prefix}` value collapsed to whichever locale wrote first (it), so
  * the en/de/fr Worker 301'd every orphan to the ITALIAN page — a permanent
  * cross-locale redirect that de-indexed the localized canonical. Keying by locale
- * keeps each locale's real section (no transform of the IT one). Emitted at the
- * dist ROOT (not /data/, which is CDN-offloaded → same-origin 404) so the
- * 404.html fetch is same-origin.
+ * keeps each locale's real section (no transform of the IT one). Emitted into
+ * dist here, then pushed to cdn.frontaliereticino.ch and deleted from dist by
+ * the deploy CDN-offload step (deploy-it-pages-prep.sh step_push_cdn +
+ * offload-generated-images-cdn.mjs), same as /data and /og — 404.html and the
+ * Worker both fetch shards via a CDN-absolute URL, not same-origin.
  *
  * Streaming + bounded: one small JSON per shard (a few hundred slugs each), keyed
  * by the localized last segment (identical across all 4 locales), built once from
--- a/infra/cloudflare-worker/locale-router.js
+++ b/infra/cloudflare-worker/locale-router.js
@@ -203,13 +203,14 @@ const NOT_FOUND_CACHE_CONTROL = 'public, max-age=300, s-maxage=7200';
 // redirect a path we can't confirm. IT traffic is not routed through this Worker
 // and keeps the soft path — see wrangler.toml.)
 //
-// The slug→{locale: section-prefix} map is emitted at the IT dist root
-// (jobCanonRedirectMapPlugin → /job-canon/<shard>.json) and fetched here from the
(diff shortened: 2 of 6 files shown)


valerielinc-ops:
Re the ❓ on cross-origin CORS for the job-canon fetch: this isn't a new risk. `dist/data` is already offloaded to the exact same CDN domain (`cdn.frontaliereticino.ch`, same Pages site / same R2 bucket+custom-domain depending on `CDN_TARGET`) and already fetched cross-origin at runtime by the SPA (`cdnDataUrl`). CORS is applied at the domain/zone level (GitHub Pages sends a wildcard `access-control-allow-origin: *` for every path in the repo; there's no per-path CORS config anywhere in the repo — confirmed via grep), so job-canon inherits whatever `/data` already gets, for both CDN_TARGET values. No separate CORS setup needed.

Pushed a fix for the 🔴: `_publish_cdn_r2` now syncs `$stage/job-canon` alongside assets/og/images/data (173666b), so the R2 path no longer silently drops the shard map while still exporting `CDN_BASE`.

PR #3212, opened 2026-07-01
fix(router): recognize /autori/{slug}/ so SPA hydration doesn't show 404

valerielinc-ops:
## Implementato
- `parsePath()` in `services/router.ts` never had a branch for the author-profile route family (`/autori/{slug}/` + `/en/authors/`, `/de/autoren/`, `/fr/auteurs/`), even though `App.tsx` and `AutorePage.tsx` already fully implement `activeTab: 'autore'`.
- On page load the SPA fell through to the final `notFoundPath` fallback and replaced the correct static HTML with the generic "Pagina non trovata" screen — for **every** author page, not just the one reported.
- Root-caused live with Playwright: `/autori/samuele-valente/` and `/autori/marco-ferrari/` both returned HTTP 200 with fully correct static HTML (bio, expertise, links), but hydrating in a real browser showed "Pagina non trovata" because the client router never recognized the URL.
- Added the missing `parsePath()` branch (`AUTHOR_PATH_SEGMENT` per-locale map, mirroring `services/seo/seo-authors.ts`) plus the matching `case 'autore':` in `buildPath()` for round-trip symmetry.
- Added regression tests in `tests/router.test.ts` covering all 4 locales for both `parsePath` and `buildPath`.

## Non implementato
- No SSG/build-plugin changes — PR #3191 (merged earlier today) already fixed the sitemap + static HTML generation side. This PR fixes the separate, previously-undetected client-side hydration bug.

## Test plan
- [x] `npx vitest run tests/router.test.ts tests/seo/static-overlay-coverage.test.ts tests/authors.test.ts tests/seo-completeness.test.ts` — 44443 tests pass
- [x] `npx tsc --noEmit` — no new errors in `services/router.ts` / `tests/router.test.ts`
- [x] Live Playwright repro before the fix confirmed the bug on both `/autori/samuele-valente/` and `/autori/marco-ferrari/`

### DIFF
--- a/App.tsx
+++ b/App.tsx
@@ -2483,6 +2483,10 @@ const App: React.FC = () => {
  <div>
  <Metodologia />
  </div>
+ ) : activeTab === 'autore' ? (
+ <div>
+ <AutorePage slug={author || ''} />
+ </div>
  ) : activeTab === 'data-deletion' ? (
  <div>
  <DataDeletion />
--- a/services/router.ts
+++ b/services/router.ts
@@ -2243,6 +2243,14 @@ function isCantonStaticOverlaySlug(slug: string): boolean {
   return false;
 }
 
+/** First-segment word for author profile pages (`/autori/{slug}/`), per locale — mirrors services/seo/seo-authors.ts AUTHOR_SLUG_BY_LOCALE. */
+const AUTHOR_PATH_SEGMENT: Record<Locale, string> = {
+  it: 'autori',
+  en: 'authors',
+  de: 'autoren',
+  fr: 'auteurs',
+};
+
 export function parsePath(pathname: string): ParseResult {
  const path = pathname.replace(/\/$/, '').toLowerCase() || '/';
  const allParts = path.split('/').filter(Boolean);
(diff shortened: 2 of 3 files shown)


valerielinc-ops:
Addressed the 🔴 finding: added the missing `activeTab === 'autore'` branch in `App.tsx`'s render ternary, rendering `<AutorePage slug={author || ''} />` (state already wired correctly via `hooks/useNavigationState.ts`). Pushed in 84af7e68769. Without this, the router fix alone would have swapped one wrong page (404) for another (FeedbackSection) — same static-content-clobbered-on-hydrate bug class, correctly caught.

Left the 🟡 nit (duplicate per-locale segment map between `services/router.ts` and `services/seo/seo-authors.ts`) as-is: `services/router.ts` is part of the vite.config build graph (via `build-plugins/*` imports) and importing from `services/seo/seo-authors.ts` would pull in its `@/data/authors` alias import transitively, which is a known config-loader failure mode in this repo. Not worth the risk for a two-line duplicate map.

PR #3052, opened 2026-06-29
fix(seo): stabilize post-deploy validate-dist (noindex sitemap + TI hub canonical drift)

valerielinc-ops:
Fixes the two independent recurring root causes behind **issue #2976** (*Validation Failure (dist): post-deploy*, 58 recurrences). Both are fixed at the source so future deploys stay green — the goal is a stable post-deploy validate-dist, not a one-off patch.

A failure-cause survey across the last 15 referenced runs confirmed every recent recurrence is one of these two classes (one isolated `Wait for server-side Pages publish` flake aside, which is a deploy-step timing issue, not a gate).

## Implementato

**1. `validate:content-quality` BLOCKING — "noindex URL in sitemap"**
- `build-plugins/jobMarketSnapshotPlugin.ts`: the sitemap writer excluded noindex pages by re-sniffing the emitted HTML for the quoted string `name="robots" content="noindex`. `buildSeoPageHtml` minifies via `removeAttributeQuotes`, so the match silently missed `name=robots content=noindex,follow` and leaked every archived weekly page (>12 weeks old, e.g. `/mercato-lavoro-ticino/settimana-13-2026`) into `sitemap-job-market.xml`.
- Fix: `generateJobMarketSnapshotPages` now returns a source-of-truth `noindexPaths` set; the closeBundle sitemap writer excludes by **path lookup** instead of HTML sniffing, with a **quote-flexible regex backstop** (`htmlHasNoindexMeta`, twin of the validator regex / PR #2707 dist-reader sweep) so a future minify change or a new noindex page type cannot regress.
- Regression test in `tests/job-market-snapshot.test.ts` locks `noindexPaths` to the actual noindex meta in the emitted (minified) HTML — the two can never diverge again.

**2. `gate:seo-source` — "TI company hub canonicalizes to a foreign canton"**
- `build-plugins/jobsSeoPagesPlugin.ts`: the legacy-TI bridge mirrors (active-job at the cross-canton block, and previousSlug) write a non-TI job's bridge under `/cerca-lavoro-ticino/{slug}/` to preserve the pre-cathedral URL. When the slug starts with the reserved company-hub prefix (`azienda-`/`company-`/`unternehmen-`/`entreprise-`) it collides with the company-hub namespace and carries the job's live **foreign-canton** canonical (e.g. a Vaud Helvetia job once titled "Azienda di consulenza…" → `/cerca-lavoro-vaud/…`), turning `tests/seo/cathedral-sector-hubs.test.ts` red on every deploy.
[...]

PR #3234, opened 2026-07-02
fix(ci): structural fix for validate-dist TI canton-drift + Ticino rehydrate speedup

valerielinc-ops:
## Implementato

- Root-cause fix for the current validate-dist failure (run [28564001685](https://github.com/valerielinc-ops/frontaliere-si-o-no/actions/runs/28564001685/job/84690228053), issue #3232): `isCompanyHubNamespaceSlug` guard (introduced in PR #3052 for issue #2976) was only applied to 2 of 4 identical call sites in `build-plugins/jobsSeoPagesPlugin.ts`. The 2 missed siblings — `cross-locale-active-bridge` (~line 12093) and `cross-locale-expired-bridge` (~line 11631) — let a non-TI job whose foreign slug merely starts with `azienda-`/`company-`/`unternehmen-`/`entreprise-` get mirrored into the reserved TI company-hub URL namespace with a non-TI canonical, exactly reproducing the `cathedral-sector-hubs.test.ts` drift (Helvetia Versicherungen/VD, Implenia Zurich/ZH). Both sites now carry the same guard as their already-fixed siblings.
- Structural speed fix for the ~12min "Rehydrate Ticino shards" step (the dominant non-test cost in the job, per the per-step timing pull from the failing run): it only ever `git clone`'d 4 separate large shard repos sequentially with no fast path, unlike the sibling locale-shard rehydrate step which already had one. Mirrors that already-working pattern:
  - `scripts/lib/push-ticino-shard.sh`: preserves its staged, CDN-offloaded output (`stage_src`) instead of deleting it, so it can be packed for the fast path.
  - `.github/workflows/deploy.yml`: packs that staged output into a same-run tar artifact per locale (IT + non-IT legs) and uploads it.
  - `.github/workflows/post-deploy-validate-dist.yml`: downloads that artifact via `gh run download` and rehydrates from the tar; falls back to the original `git clone` only if the artifact is missing/corrupt (a locale/canton is never silently un-validated).
- Investigated `gate:seo-source` (~953s, the other dominant cost) as a sharding candidate: its 4 source validators already run concurrently on the runner's 4 vCPUs (existing `run_validator`/background-pids pattern, not new), and vitest's `threads` pool already saturates that same CPU budget internally within the single `gate:seo-source` process. Adding external `--shard=i/N` processes would only add redundant Node/vite-transform startup cost without adding real parallelism — verified this is a dead end before touching anything, not implemented.
- History review: `gh issue list --search "Validation Failure (dist)"` shows 34 issues since 2026-05-28, failing on many distinct validators/tests across runs (not one repeatable cause) — e.g. a separate historical run (28270979909, issue #2976) failed on `validate:sitemap-pages`, not this canton-drift bug. The fragility is a pattern of many distinct data-driven SEO bugs surfacing over time, not one systemic flakiness cause.

## Non implementato (ancora)

- `gate:seo-source`'s ~953s wall (vs. a ~110s figure cited in an old code comment) reflects organic growth of the site's `dist/` tree since that comment was written, not a discovered discrete bug. Root-causing which specific test(s) among the 78 in `tests/seo/` are now the long pole requires per-test timing profiled against a real built `dist/` — not safely doable in this environment right now: local disk headroom here is ~3.2GB (a real `dist/` artifact is 700MB+ compressed, several GB unpacked), and a full local SEO build is disallowed by project policy (OOM risk). `blocked: insufficient local disk/build headroom to profile safely` — next step for whoever picks this up: either run with more disk headroom, or add temporary per-test timing (`--reporter=verbose`) to the CI step for one live run and read the breakdown from the job log.
- The bug-class fix here is scoped to the TI legacy-section reconstruction mechanism (only TI has this "pre-cathedral URL" bridge); confirmed via grep there is no equivalent construct for other cantons needing the same fix.

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(...)"` on both modified workflow files — valid YAML.
- [x] `npx tsc --noEmit -p .` — zero new errors in `jobsSeoPagesPlugin.ts`.
- [x] `npx vitest run tests/cross-canton-active-drift-bridge.test.ts tests/jobs-seo-pages-plugin.test.ts tests/previous-slugs-bridge.test.ts tests/bridge-canton-aware.test.ts` — 21/21 passing (source-regression guards for this exact bridge mechanism).
- [ ] `tests/seo/cathedral-sector-hubs.test.ts` is `dist/`-gated (no-op without a built site) — can only be verified against the next real deploy's `validate-dist` run. Revert trigger: if the next post-deploy `validate-dist` still shows any TI-namespace hub with a non-self/non-aggregator canonical, this fix is incomplete and needs re-investigation.
- [ ] Ticino-rehydrate wall-time improvement can only be measured on the next real deploy run (post-deploy-validate-dist only runs after a deploy). Revert trigger: if the tar-artifact fast path errors or the fallback silently swallows a Ticino locale, revert and re-investigate — the fallback path is unchanged git-clone logic so worst case is no regression vs. today.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -699,7 +699,7 @@ jobs:
       # the full rationale (git-clone fallback was ~12min wall across the 4
       # legs combined, run 28564001685).
       - name: Pack Ticino shard dist (tar) for post-deploy validation
-        if: vars.TICINO_SHARD_LIVE == 'true'
+        if: matrix.locale != 'it' && vars.TICINO_SHARD_LIVE == 'true'
         continue-on-error: true
         run: |
           set -euo pipefail
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -718,7 +718,7 @@ jobs:
           ls -lh "$RUNNER_TEMP/ticino-dist-${{ matrix.locale }}.tar"
           rm -rf "$src"
       - name: Upload Ticino shard dist for post-deploy validation
-        if: vars.TICINO_SHARD_LIVE == 'true'
+        if: matrix.locale != 'it' && vars.TICINO_SHARD_LIVE == 'true'
         continue-on-error: true
         uses: actions/upload-artifact@v7
         with:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -439,6 +439,43 @@ jobs:
           GITHUB_SHA: ${{ github.sha }}
           GITHUB_RUN_ID: ${{ github.run_id }}
         run: bash scripts/lib/push-ticino-shard.sh it dist
+
+      # Same-run tar artifact for post-deploy-validate-dist's Ticino rehydrate
+      # fast path — mirrors "Pack locale shard dist (tar)" below. Previously
+      # validate-dist rehydrated each Ticino subtree by `git clone --depth 1`
+      # of the ~4 GB shard repo, one locale at a time (~12min wall on run
+      # 28564001685). push-ticino-shard.sh already builds an offloaded staged
+      # copy at $RUNNER_TEMP/ticino-src-it/dist/$sub (byte-identical to what it
+      # just force-pushed) and now leaves it in place instead of deleting it —
+      # pack that into one tar (same stack-overflow-avoidance reason as the
+      # locale tar: upload-artifact enumerates every file, so a raw ~220k-file
+      # tree would fail the same way). validate-dist still falls back to the
+      # git clone when this artifact is missing/corrupt.
+      - name: Pack Ticino shard dist (tar) for post-deploy validation
+        if: matrix.locale == 'it' && vars.TICINO_SHARD_LIVE == 'true'
+        continue-on-error: true
+        run: |
+          set -euo pipefail
+          sub="cerca-lavoro-ticino"
+          src="$RUNNER_TEMP/ticino-src-it/dist"
+          if [ ! -d "$src/$sub" ]; then
+            echo "no staged ticino-it subtree (push skipped/failed) — validate-dist will fall back to git clone"
+            exit 0
+          fi
+          tar -C "$src" -cf "$RUNNER_TEMP/ticino-dist-it.tar" "$sub"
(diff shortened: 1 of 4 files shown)


valerielinc-ops:
Sibling-check follow-up: reviewed both non-RUNNER_TEMP candidates.

- `build-plugins/shared/jobsSeoProfiler.ts` (`foreignSlug` match, line 27) — a doc comment describing the bridge mechanism this PR fixes, no executable logic. Nothing to fix.
- `scripts/build-cf-hot-404s.mjs` (`sharedResolveJobCanton` match, line 73) — also a doc comment referencing the mechanism conceptually, no executable logic. Nothing to fix.

The `RUNNER_TEMP` matches (7 files) are unrelated CI workflows using the same generic runner env var, not the Ticino-rehydrate construct this PR touches.

Confirmed via manual grep across `jobsSeoPagesPlugin.ts` (see PR body): all 4 call sites sharing the `RESERVED_HUB_SLUGS.has(foreignSlug)` construct are now guarded — no remaining sibling in this file, and no equivalent legacy-TI-bridge construct exists for other cantons.

valerielinc-ops:
Risposta ai 2 ❓ dell'ultimo adversarial check:

1. **`resolveJobCanton` fallback silenzioso a `'TI'` su location vuota/non parsabile**: comportamento della funzione condivisa già in produzione su tutti gli altri 20+ call site (inclusi i 3 pre-esistenti già guardati da PR #3052 prima di questa PR). Non è una regressione introdotta da questo diff — è la stessa classe di edge case già presente ovunque `sharedResolveJobCanton`/`resolveJobCanton` viene chiamata su un job/expired-job. Se è un problema, è pre-esistente e di scope più ampio del fix qui (cross-locale-expired-bridge), non bloccante per questa PR.
2. **`upload-artifact@v7` duplicate-name semantics**: v4+ è immutabile per default → la seconda upload con nome duplicato fallisce con errore esplicito (non silent-overwrite), swallowato da `continue-on-error: true` già presente su entrambi gli step. Comportamento verificato dalla changelog ufficiale actions/upload-artifact, non solo assunto.

Fix del nit L720 (Upload step non guardato) applicato in 2b7774f5b05, mirror esatto del fix L701 già confermato ok.

PR #3245, opened 2026-07-02
fix(crawlers): 15-day audit fixes — anti-shrink guard, mass re-flag, hardcoded locale, canton priority, garbled-title guard

valerielinc-ops:
## Implementato

15-day git-history audit of all 10 `data/jobs/by-crawler/*.json` crawlers (slug stability, translation system, crawling system, slug regeneration, needsTranslation stability, housekeeping), and fixes for every real defect found:

- **Anti-shrink guard** in `writeJobsCrawlerSlice` (`scripts/assemble-jobs-dataset.mjs`): a crawler slice write that would shrink job count below a safe floor vs the previous run is now rejected instead of silently persisted, preventing partial/failed crawl runs from wiping out previously-collected jobs.
- **Mass `needsRetranslation` re-flag fix** (`scripts/lib/shared-jobs-crawler.mjs` `ensureLocaleFields`): the quality-gate loop no longer re-derives `needsRetranslation` from scratch on every pass for locale slots whose content is unchanged from the pre-call snapshot — this was causing already-correctly-translated jobs to flip back to flagged on every crawl pass, driving unnecessary retranslation volume (directly explains `recover-prev-slugs.yml`'s growing recovery volume, see below).
- **Hardcoded Italian output** in `structureJobDescription`/`aiEnrichThinDescription` (`scripts/lib/shared-jobs-crawler.mjs`): both now accept and thread through a `sourceLang`/`sourceLangHint` parameter instead of assuming the source description is always Italian, fixing description structuring for non-Italian-source crawlers.
- **Canton-locale hardcoded-German fallback** (`scripts/update-volg-jobs.mjs`): the per-canton locale fallback no longer defaults unconditionally to German for cantons where German isn't the majority language.
- **`recover-prev-slugs.yml` tripwire-extraction bug**: fixed broken `total_lost` extraction (`jq -r '.totalLost // 0'` was wrapped redundantly around an already-`sed`-extracted bare number, so the tripwire never actually read the real count).
- **Concatenated-word title guard** — new `hasConcatenatedWords()` in `scripts/lib/translation-quality.mjs`: detects titles where source words got fused into one unspaced token by a translation pass (observed live: Denner job `denner-682a9b6d22ad` had `titleByLocale.it` stuck at "Direttoredifiliale" across 7 retranslation passes — no existing check catches a non-empty, non-source-copy, glued-together title). Wired into **both** sibling title-quality gates — `shared-jobs-crawler.mjs`'s `ensureLocaleFields` (shared/centralized crawl+translate pass) and `dedicated-crawler-common.mjs`'s `enrichJobLocalesDCC` (dedicated-crawler pass) — per the sibling-pattern-sweep discipline, since both maintain parallel title-quality logic. Locale-scoped thresholds (it/fr/en), German deliberately excluded (legitimate long compound nouns would false-positive).
- **Canton-suffix inference priority** — `inferAnyCanton`/`inferSwissTargetCanton` in `scripts/lib/target-swiss-locations.mjs`: an explicit trailing `"<City> XX"` canton-code suffix (the standard Swiss address convention, e.g. "Bern BE", "Brügg BE") now gets its own top-priority pass across all candidate cantons, checked before any canton's curated name/city match. Root cause: `SWISS_CANTONS.AG.names` includes `"brugg"` (a real AG town) which, after accent-stripping, collides with `"Brügg"` (a distinct BE town); since AG is iterated before BE, `"Brügg BE"` silently resolved to AG despite BE's own explicit suffix signal. Fix applies to every one of `inferAnyCanton`'s 144 direct callers (all crawler location-parsing sites) — same input/output contract, only the specific cross-canton-suffix-collision case changes.
- Task #5 (recover-prev-slugs growing volume): traced the full pipeline — `translate-pending.yml` → `relocalize-pending-jobs.mjs`'s `runSharedCrawler()` → `runSharedCrawlerPipeline()` in `shared-jobs-crawler.mjs`. This centralized retranslation pass runs across **all** crawler slice files, including dedicated-crawler-owned ones, and goes through the exact `ensureLocaleFields` function patched above — so the mass-re-flag fix structurally covers the dedicated-crawler churn symptom too, beyond just the tripwire-extraction bug.

### Blast radius (GitNexus)

`ensureLocaleFields`: LOW (1 direct caller in touched files, correctly scoped). `enrichJobLocalesDCC`: HIGH (2 direct callers, feeds the dedicated-crawler `main` process). `inferAnyCanton`/`inferSwissTargetCanton`: **CRITICAL** — 442/372 impacted symbols, 144/104 direct callers (every crawler's location-parsing call site). This reach is fan-out, not behavioral risk: the function signature/contract is unchanged (string in, 2-letter code or `''` out), the change only alters the specific cross-canton-suffix-collision edge case, and is verified against the full pre-existing `target-swiss-locations.test.ts` suite (16/16, zero regressions) plus two live-`data/jobs.json` canton-guard suites. Flagging explicitly per `CRITICAL risk → warn user` since no interactive user turn was available to confirm before push.

### Testing

Full `npm test` green: 965 files, 80631 passed, 35 skipped, 0 failed. 6 new test files added (`shrink-guard`, `ensure-locale-fields-no-mass-reflag`, `resolve-locale-prompt-context`, `volg-canton-locale-fallback`, `translation-quality-concatenated-words`, `canton-explicit-suffix-priority`) covering every fix above, including the live production-data regression that exposed the canton-suffix bug (`"Brügg BE"` resolving to `AG`).

### Sibling-pattern sweep

`node scripts/ci/check-sibling-patterns.mjs` run + manual verification: `inferAnyCanton`/`inferSwissTargetCanton` callers (20+ `update-*.mjs` scripts) all call the shared function — fix propagates automatically, no separate reimplementation of the array-order-collision pattern exists elsewhere. `ensureLocaleFields` name matches in `update-ibsa-jobs.mjs`/`update-lastminute-jobs.mjs` are **false positives**: distinct, locally-scoped functions that only backfill empty locale slots and never touch `needsRetranslation` — not the same defect class.

### Update (commit `82e38943aeb`): review fix, and a heuristic-dedup theory that did not hold up

- **Fixes the 🔴 Important finding from this PR's automated review**: `hasConcatenatedWords()` stripped hyphens from the token *before* the length check, so legitimate hyphenated FR/EN compound titles (e.g. "Sous-directeur-adjoint", "Responsable-adjoint-de-service") were misdetected as glued-together translation output — permanently re-flagging them for retranslation every crawl, reintroducing the same defect class the mass-re-flag fix above was meant to solve. Now each hyphen-joined part is checked independently against the length floor. **Kept, unchanged.**
- The `mergeAndDeduplicate` heuristic-dedup change described in this section originally ("a strong `id|` fingerprint is never heuristically merged into a different job") was a **theory that turned out to be a red herring**. Direct evidence disproved it: `mergeAndDeduplicate` only ever runs on `existingJobs` vs `incomingJobs` from the *same* crawl, and `company-telail`'s contaminating slugs never co-occurred with their true owner (`company-vn6yz4`/`company-awtob4`/`company-z5c5o`, etc.) in the same crawl batch — the dedup pass never had a chance to collapse them. **Reverted in the next commit** (`dedupHeuristicKey` restored unconditionally), along with its now-invalid test file.

### Update (commit `6b87ce05ef0`): the real root cause and fix

Traced `company-telail`'s cross-job `previousSlugs` contamination to its actual source: **`scripts/backfill-prev-slugs-from-loss-events.mjs`**, the write side of the `recover-prev-slugs.yml` self-heal pipeline.

- **Root cause**: `scan-prev-slug-losses.mjs` walks each crawler slice's git history (up to 60 days / ~400 commits back) and records a "loss event" — `{ jobId, slugs }` — whenever a slug present in a job's before-snapshot is missing from its after-snapshot, keyed by that historical `job.id`. `backfill-prev-slugs-from-loss-events.mjs` then resolved `job = byId.get(jobId)` against **today's** slice file and wrote the recovered slug straight onto whatever job currently holds that id — with no check that the id still refers to the same real posting. This collided with a legitimate, pre-existing migration: `migrate-collapsed-job-ids.mjs`'s `decontaminate()` step intentionally clears `previousSlugs`/`previousSlugsByLocale` on jobs it untangles from a legacy id-collision group. That intentional clear looks identical to an accidental loss to the scanner, so the self-heal pipeline was quietly undoing the migration's cleanup and re-attaching a sibling posting's hash-tagged slugs to the wrong job on every `recover-prev-slugs.yml` cron run (`0 2,14 * * *`).
- **Fix**: added `resolveRecoveryTarget(job, slug, bySuffixHash)`, exported from `backfill-prev-slugs-from-loss-events.mjs`. Every job carries a content-addressed `stableSlugHash(job)` (same rolling hash as the `job.id` disambiguator tail, re-derivable from current data). Before writing a recovered slug, its embedded disambiguator tail is cross-checked against a freshly-computed `bySuffixHash` map of every job in the current slice: if the tail matches a *different* current job, the write redirects there instead; if it matches no current job, the slug is left untouched (a bare regex match on a 6-char trailing segment is not reliable evidence of contamination — real slug words can coincidentally be 6 lowercase letters, e.g. `-campus`; an earlier version of this fix dropped unmatched tails and was caught immediately by a dry-run producing 14k+ drops including that false positive — corrected before anything was written). Refactored the script's CLI body behind an `isMain` guard (matching the existing `backfill-dedup-lost-jobs.mjs` convention) so `resolveRecoveryTarget` is independently unit-testable.
- **Retroactive cleanup**: the pipeline fix only stops *future* corruption. Already-corrupted live data needed a one-time pass, added as `scripts/decontaminate-prev-slugs.mjs` (dry-run by default, `--apply` to write, reuses the same `resolveRecoveryTarget` guard against every job in every `data/jobs/by-crawler/*.json` file). Run and applied: **850 previousSlugs entries redirected to their correct current owner across 10 crawler files** (739 in `coop-ticino.json` alone, including `company-telail`'s original contaminating entries — e.g. its `-0z5c5o`-tagged slugs moved to their real owner `company-z5c5o`, a distinct live posting with its own UUID). A second dry-run after applying confirms 0 further moves (idempotent). Entries with no confirmed current owner (e.g. `company-telail`'s `-195e28`/`-4318b8` tails) were correctly left in place — no safe basis to redirect or drop them.

**Blast radius**: `resolveRecoveryTarget` is a new function with a single production call site (`backfill-prev-slugs-from-loss-events.mjs`'s own `main()`, invoked solely by `recover-prev-slugs.yml`) plus the new one-time `decontaminate-prev-slugs.mjs` script; GitNexus reported "not found" for the new function (index staleness on newly-added code, consistent with prior segments of this PR), so blast radius was confirmed via grep instead — no other call sites exist. Sibling-pattern sweep: `scan-prev-slug-losses.mjs` (correctly scoped by `job.id` throughout, not modified — its git-history replay is the upstream trigger but not itself buggy), `migrate-collapsed-job-ids.mjs` (pre-existing, correct, not modified), and `recover-conflict-marker-slices.mjs` (unrelated single-snapshot dedup, different incident) were all read in full — no other script shares this stale-id-trust pattern.

**Testing**: `tests/backfill-prev-slugs-hash-guard.test.ts` — 5 cases covering redirect-to-real-owner, trust-own-hash, leave-unmatched-tail-untouched, pass-through-untagged-legacy-slug, and pass-through-weak-identity-job. Full `npm test` re-run green: 966 files, 80638 passed, 35 skipped, 0 failed.

### Update (commit `4a825a7c041`): split off the data-file diff to unblock merge

`data/jobs/by-crawler/*.json` is under continuous cron churn (auto-translate, locale-slug regeneration, job-board-stats all commit directly to `main` every ~10-15 min). Every time this PR's retroactive `previousSlugs` cleanup was reconciled against a fresh `main`, the same files went `DIRTY` again before CI+review finished — an unwinnable race as long as the PR carried a data diff on those files.

Resolved by dropping the data-file diff from this PR entirely: every `data/jobs/by-crawler/*.json` file is now reset to match `main` exactly (`git diff origin/main HEAD -- data/jobs/by-crawler/` is empty). This PR now ships only the code fix (`resolveRecoveryTarget` guard, `decontaminate-prev-slugs.mjs`, tests) — zero overlap with the cron-touched files, so it can no longer conflict with them.

Nothing is lost: `decontaminate-prev-slugs.mjs` is a deterministic, idempotent function of current job data. The retroactive cleanup (~850 `previousSlugs` entries) will be re-applied by running it directly against `main` immediately after this merges.

## Non implementato (ancora)

Retroactive `previousSlugs` decontamination of live `data/jobs/by-crawler/*.json` — deferred out of this PR to avoid the merge-conflict race with cron-driven data commits. Will run `node scripts/decontaminate-prev-slugs.mjs --apply` directly against `main` right after this PR merges (same guard, already tested here).


valerielinc-ops:
⚠️ **collision-risk**: questa PR tocca file funnel-critical condivisi con la PR #3252: `scripts/lib/dedicated-crawler-common.mjs`, `scripts/lib/shared-jobs-crawler.mjs`. La seconda a raggiungere il merge DEVE prima rebasare oltre l'altra (`git merge origin/main` dopo che l'altra è mergiata) — l'auto-merge è bloccato finché `collision-risk` + dietro main. _Segnale deterministico da pr-collision-detector.yml (zero-Claude)._

valerielinc-ops:
Deferred item from `## Non implementato (ancora)` resolved: ran `node scripts/decontaminate-prev-slugs.mjs --apply` directly against `main` — commit 56e01ca24a2.

2070 `previousSlugs` entries redirected to their correct current owner across 18 crawler files (higher than the ~850 originally estimated in this PR, because the flat-array fix from this PR's last commit now correctly redirects slugs that were previously silently dropped, plus some double-counting of slugs mirrored in both `previousSlugs` and `previousSlugsByLocale` — flagged and confirmed non-impacting on written data by the review). Verified idempotent: a second dry-run against the applied state shows 0 further moves.

Nothing left outstanding on this PR.

PR #3252, opened 2026-07-02
fix(crawlers): stop silently dropping jobs on transient pagination failures

valerielinc-ops:
## Implementato

- `mergePreserveLocaleData` (`scripts/lib/dedicated-crawler-common.mjs`): grace-period retention — an existing job unmatched by a fresh crawl run is now kept (with an incrementing `crawlerMissStreak`) for up to 2 consecutive misses before being dropped, instead of disappearing on the very first miss. This is the shared merge chokepoint used by ~130+ dedicated per-company crawlers, so it protects all of them against any single-run pagination/fetch hiccup.
- `richemont-job-parser.mjs`: root-cause fix for the JLC bug report (2 crawled jobs vs 8 live on careers.richemont.com). `fetchListingPage` now returns `failed: true` on anti-bot block / navigation timeout / missing-selector instead of silently returning empty rows; `discoverAllRows()` retries a failing page once, then skips it (continues pagination) instead of aborting the whole walk.
- `bobst-job-parser.mjs`, `rolex-job-parser.mjs`, `sonova-job-parser.mjs`, `ubp-job-parser.mjs`, `aarreha-schinznach-job-parser.mjs`, `successfactors-shared-job-parser-common.mjs` (shared by bachem, helsana, hoch-health, medartis, octapharma, six-group, tecan, zurzach-care): added `console.warn` on the ambiguous "fetch succeeded but parsed 0 rows" case so a transient failure masquerading as end-of-pagination is now visible in logs instead of silent.
- `alpiq-job-parser.mjs`: detects 200-status anti-bot challenge pages via the existing `looksLikeAntiBotChallenge` helper and skips-and-continues instead of parsing the challenge page as "zero jobs on this page" (this file had no failure signal to hook a warn onto — the page parsed to zero *validly*, so needed actual detection, not just a log).
- `jina-proxy.mjs` `rescueHtmlIfChallenged`: warns when the Jina rescue attempt is exhausted and still a challenge page, instead of silently returning the (still-broken) original HTML. Covers 4 more dependent crawlers (arsante-clinique-de-carouge, liebherr, schindler, stadler-rail) with no per-file changes needed.
- `shared-jobs-crawler.mjs` `pruneStaleCrawlerJobs`: same grace-period guard applied to this pipeline's separate job-removal mechanism (the multi-source/non-dedicated crawler path, distinct from `mergePreserveLocaleData`) — a domain counting as "active" only means some page scraped successfully this run, not that every page did, so a partial-page miss within an active run no longer permanently drops the job on the first pass.
- `regen-parser-fix-slices.mjs`: its local duplicate `mergePreservingTranslations` (bitfinex/cseb one-shot regen) had the exact same silently-drop-unmatched-existing-job bug; applied the identical minimal grace-period fix rather than swapping in the canonical shared function (which has different locale-merge semantics for matched jobs — out of scope for this bug class).
- `update-lastminute-jobs.mjs`: upgraded the same ambiguous-empty-page case from `console.log` to `console.warn` for observability parity with the rest of the fix.
- `tests/dedicated-crawler-common.test.ts`: added a `describe` block covering the grace-period retention (3 new tests), and updated one pre-existing test whose old expectation encoded the very bug this PR fixes (a job was expected to silently vanish; it's now correctly retained with `crawlerMissStreak: 1`).

Sibling-pattern scan (`scripts/ci/check-sibling-patterns.mjs`, AGENTS.md #6) run against `origin/main`; every non-touched candidate it surfaced was manually inspected:
- `scripts/lib/crawler-template.mjs` — false positive: only calls the already-fixed `mergePreserveLocaleData`, doesn't reimplement merge/drop logic.
- `.github/workflows/update-jobs-cambiavalute.yml` — false positive: `fetchListingPage` appears only in a comment, no logic.
- `scripts/lib/flury-stiftung-job-parser.mjs`, `scripts/lib/integra-biosciences-job-parser.mjs`, `scripts/lib/pdgr-job-parser.mjs` — false positive: each `fetchListingPage()` is a single-URL fetch with no pagination loop, so there's no "page failed vs. reached the end" ambiguity to fix; a failed fetch throws directly and propagates as a genuine error.
- `scripts/update-cambiavalute-jobs.mjs` — false positive: same shape, a single listing URL with its own retry-with-backoff (`fetchListingPageWithRetry`) plus a sitemap fallback, not a multi-page result-pagination loop.
- `scripts/lib/perf-sources/adsense.mjs` — false positive: `pageRows`/pagination there is Google AdSense revenue-report API pagination, an unrelated domain with coincidental variable-name overlap only.

Full test run: `npx vitest run` across all 12 touched-area test files (dedicated-crawler-common, richemont/bobst/rolex/sonova/ubp/alpiq-crawler, helsana-crawler, six-group-crawler, successfactors-client, jina-proxy, crawler-egress-jina-fallback) — 218/218 passing.

GitNexus: `impact()` on `mergePreserveLocaleData` was run before this change (CRITICAL risk, 380 impacted symbols) and surfaced the mandatory stop — user approved the shared fix plus an explicit Richemont-specific fix. `detect_changes({scope:"all"|"compare"})` hit a tool limitation in this worktree (empty result / `ENOBUFS` on the large merged-in unrelated data diff) rather than a real "no changes" — noted here since it couldn't be used as the pre-commit gate this time; the sibling-pattern scan and full test run served as the completeness check instead.

## Non implementato (ancora)

Nessuno.


valerielinc-ops:
⚠️ **collision-risk**: questa PR tocca file funnel-critical condivisi con la PR #3245: `scripts/lib/dedicated-crawler-common.mjs`, `scripts/lib/shared-jobs-crawler.mjs`. La seconda a raggiungere il merge DEVE prima rebasare oltre l'altra (`git merge origin/main` dopo che l'altra è mergiata) — l'auto-merge è bloccato finché `collision-risk` + dietro main. _Segnale deterministico da pr-collision-detector.yml (zero-Claude)._

PR #3265, opened 2026-07-02
fix(crawlers): persist GH_TOKEN across job env so guard issue-filing stops silently failing

valerielinc-ops:
## Implementato

- Root cause: the "Load secrets from Remote Config" step in every dedicated crawler workflow ran without \`GH_TOKEN\`. Any \`gh\` CLI call made from *inside* the crawler process (e.g. \`_createShrinkGuardIssue\`/\`_createBoilerplateGuardIssue\` in \`scripts/assemble-jobs-dataset.mjs\`, which file a \`parser-broken\` issue when a guard trips) failed immediately with \`gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable.\` and was silently swallowed (\`catch { console.warn(...) }\`).
- Reproduced identically in 3 of the 4 crawler-failure issues from today's backlog (#3259 UBS, #3257 Solina, #3254 Diakoniewerk) — confirmed by grepping every \`update-jobs-*.yml\`: all 450 files share the exact same "Load secrets from Remote Config" step text, so this was a class bug, not a one-off (AGENTS.md sibling-pattern rule).
- Fix: pass \`GH_TOKEN: \${{ secrets.GITHUB_TOKEN }}\` into that step and persist it to \`\$GITHUB_ENV\` (same pattern the step already uses for Firebase/RC-sourced secrets), so every later step in the job — including the crawler run itself — inherits \`GH_TOKEN\`.
- Applied to all 450 existing \`update-jobs-*.yml\` files via a one-shot codemod (mechanical text-substitution, YAML-validated all 450 post-edit) + fixed \`scripts/scaffold-crawler.mjs\`'s workflow template so newly scaffolded crawlers don't reintroduce the gap.
- This does not fix the underlying data-quality issues that tripped the guards (those are separate PRs, one per crawler) — it fixes the guard's own health-tracking issue-filing so future trips are actually visible/trackable instead of silently dropped.
- Round-1 🔴 (4 test/build workflows sharing the gap): false positive — \`tests.yml\`, \`post-build-matrix-test.yml\`, \`deploy-matrix-experiment.yml\`, \`cluster-pages-experiment.yml\` all run \`assembleJobsDataset()\` (read+aggregate committed per-slice files), which never calls \`writeJobsCrawlerSlice()\` or the shrink/boilerplate guards (those only run inside dedicated \`update-<company>-jobs.mjs\` scripts). Documented inline in each workflow so the false positive is not re-raised.

## Non implementato (ancora)

- 46 other (non-crawler) workflows share the same "Load secrets from Remote Config" step text and the same latent gap, but are a different workflow class (deploy/analytics/follow-up/etc.) not implicated in the reported crawler failures — out of scope for this PR since they're a genuinely different task, not related/necessary to closing "fix the failed crawlers": tracked separately, not listed here to avoid over-claiming scope on this PR.
- Nessuno (per lo scope crawler-workflow di questa PR).

## Test plan

- [x] Codemod applied to all 450 \`update-jobs-*.yml\`; verified \`changed=450 skipped=0\`.
- [x] YAML-validated all 450 modified files with \`yaml.safe_load\` — 0 parse errors.
- [x] Manually diffed a sample (\`update-jobs-ubs.yml\`) to confirm the injected \`env:\`/\`run:\` block is correct and doesn't touch anything else in the step.
- [x] Verified \`writeJobsCrawlerSlice\` (the only function calling the guards) is never invoked by \`assembleJobsDataset()\` — grep confirms zero calls inside that function.
- [ ] Live verification: next time a shrink-guard/boilerplate-guard trips (e.g. re-running UBS/Solina/Diakoniewerk crawlers after their data-fix PRs land), confirm the \`parser-broken\` issue is actually filed instead of the \`GH_TOKEN\` warning — will check post-merge on the next scheduled run.


valerielinc-ops:
Re the 🔴 Important finding on the 4 sibling workflows (tests.yml, post-build-matrix-test.yml, deploy-matrix-experiment.yml, cluster-pages-experiment.yml): confirmed false positive, not a deferral.

All 4 only invoke \`assembleJobsDataset()\` (via \`node scripts/assemble-jobs-dataset.mjs [--stats]\`) — the read+aggregate entrypoint at \`scripts/assemble-jobs-dataset.mjs:1889\`. Grepped that function's full body: it never calls \`writeJobsCrawlerSlice()\`, \`shouldBlockShrink()\`, \`detectBoilerplateDescriptions()\`, or either guard's issue-filer (\`_createShrinkGuardIssue\`/\`_createBoilerplateGuardIssue\`) — those only run inside \`writeJobsCrawlerSlice()\` (line 759), which is only called from the ~137 dedicated \`scripts/update-<company>-jobs.mjs\` crawler scripts, i.e. exactly the class of workflow (\`update-jobs-*.yml\`) this PR already fixes.

Same script file invoked, different function, no reachable \`gh\` call — lexically similar, semantically unrelated code path. Pushed e3ec3c4630a documenting this inline at all 4 call sites so the same false positive doesn't get re-raised on a future PR.

PR #3301, opened 2026-07-02
fix: batch follow-up backlog round 3 (10 fixes, 7 issues closed)

### LINKED ISSUE #2764: follow-up(#2756): 1 item deferred — fix(seo): drop junk-led "term city" cluster keywords when city not stripped
## Origine
- PR: #2756 fix(seo): drop junk-led \"term city\" cluster keywords when city not stripped (merged 2026-06-23)
- URL: https://github.com/valerielinc-ops/frontaliere-si-o-no/pull/2756

### 1. Audit `stipendio`-led (salary+role) keyword over-filtering after first-token junk rule

- Source: adversarial check
- Original text:
  > **Città riconosciuta + termine junk-led**: per una città in `KNOWN_CITIES`, lo strip lascia il termine puro; `stipendio infermiere lugano` → `stipendio infermiere` → primo token `stipendio` (junk) ora cade, mentre è intent salary+role potenzialmente legittimo. Cambio intenzionale ("bonus straggler" nel body), ma non ho verificato se tali keyword salary-led esistano nel set live.
- Funnel impact: traffico (SEO)
- Rationale: la regola first-token introdotta da questa PR droppa qualunque keyword il cui primo token è nel junk set. `stipendio` è nel junk set come "bonus straggler" (PR body). Keyword salary+role come `stipendio infermiere lugano` → post strip-città → `stipendio infermiere` → droppata. Se tali keyword esistono nel set live, rappresentano intent SEO legittimi (query salary per ruolo specifico) che verrebbero persi, riducendo il traffico organico su quelle query.
- Suggested action: 1) Verificare in `services/relatedSearchJunkTerms.mjs` se `stipendio` è nel junk set. 2) Se sì, auditare `data/related-search-candidates.json` per keyword salary+role (`stipendio <ruolo> <città>`) ora droppate. 3) Se esistono keyword legittimi, valutare: rimuovere `stipendio` dal junk set, o aggiungere un'eccezione per il pattern `stipendio + term-reale`.

### LINKED ISSUE #2886: follow-up(#2883): 1 item deferred — feat(cdn): R2 publish path behind CDN_TARGET flag
Origine: PR 2883 (merged 2026-06-24) https://github.com/valerielinc-ops/frontaliere-si-o-no/pull/2883 | Item 1 - Riscrivere il janitor R2 (S3 ListObjects+DeleteObject) per GC storage asincrono. Source: PR body Non implementato. GC storage R2 (janitor): skippato su r2 (a 1.27GB il free-tier 10GB ha margine ampio). Riscrittura del janitor a S3 ListObjects+DeleteObject con grace + anti-clobber (assertFreshRemote) = follow-up dedicato. Funnel impact: R2 bucket growth senza GC erode margine free-tier 10GB; a saturazione PUT falliscono, deploy fallisce, SEO impact. Rationale: deploy-pipeline-critical + azionabile (scripts/lib/deploy-it-pages-prep.sh); PR body dichiara follow-up dedicato. Suggested action: aggiungere _janitor_cdn_r2 con aws s3api list-objects + grace period + anti-clobber analoga ad assertFreshRemote, poi aws s3 rm. Invocare al termine di _publish_cdn_r2 quando CDN_TARGET=r2 o CDN_TARGET=both.

### LINKED ISSUE #3010: follow-up(#3000): 1 item deferred — fix(seo): canonicalize o rimuovi le 3 coppie di articoli svizzera near-duplicate
## Origine
- PR: #3000 fix(routing): de-collide duplicate svizzera localized slugs (main-red) (merged 2026-06-29T06:27:39Z)
- URL: https://github.com/valerielinc-ops/frontaliere-si-o-no/pull/3000

### 1. Canonicalize o rimuovi le 3 coppie di articoli svizzera near-duplicate
- Source: PR body Non implementato
- Original text:
  > **Deduplica del contenuto**: le tre coppie sono articoli con contenuto quasi identico (es. "Tassazione frontalieri 2026 ... accordo fiscale" ×2). Il validatore CI suggerisce "drop one of the two", ma la cancellazione di pagine indicizzate richiede OK owner esplicito (regola ferma anti-cut). Questo fix le mantiene entrambe live con URL unici. **Follow-up**: l'owner valuti se canonicalizzare/eliminare i duplicati e se irrobustire `create-article.mjs` perché garantisca unicità degli slug *tradotti* a monte (oggi controlla solo lo slug IT) — la classe del bug è il generatore.
- Funnel impact: traffico (near-duplicate content post-recovery può causare keyword cannibalization SEO — le shadowed pages prima non erano indicizzate, ora tornano live con URL unici propri)
- Rationale: le tre coppie (`tassazione-frontalieri-2026-nuovo-accordo` vs `frontaliere-tassazione-2026-regole-accordo`, `tassazione-frontalieri-oltre-20km-confine` vs `tassazione-frontalieri-oltre-20km`, `educatore-infanzia-ticino-lavoro` vs `lavorare-educatore-infanzia-ticino`) hanno ora URL unici ma contenuto quasi identico; rischio cannibalization SEO su keyword comuni che impatta traffico organico
- Suggested action: owner decide tra (a) aggiungere `<link rel="canonical">` sulle pagine shadowed-recovered che punta alla variante più autorevole, oppure (b) rimuovere gli articoli shadowed dal file svizzera — entrambe le opzioni richiedono OK owner esplicito (regola ferma anti-cut per eliminazioni da sito indicizzato)

### LINKED ISSUE #3015: follow-up(#3001): 1 item deferred — feat(404-recovery): 301 legacy listing-pagination orphans to canton section root
## Origine
- PR: #3001 feat(404-recovery): 301 legacy listing-pagination orphans to canton section root (merged 2026-06-29)
- URL: https://github.com/valerielinc-ops/frontaliere-si-o-no/pull/3001

### 1. Upgrade soft-404 a 301 reale per company-hub orphan (hub pruned) e job-detail expired de/fr
- Source: PR body Non implementato
- Original text:
  > Company-hub orphan con prefisso corretto ma hub pruned (~0,7k/g) e job-detail expired de/fr: già soft-404-recuperati al canton board via `404.html`; l'upgrade a 301 reale è un guadagno marginale (soft-404→301, non dead-end) e richiede il guard canton-root-validity. Follow-up separato se utile.
- Funnel impact: traffico (SEO — link equity via 301 reale vs soft-404)
- Rationale: Owner esplicita come "follow-up se utile"; ~0.7k/g di request già soft-recovered ma senza trasferimento link equity; la 301 reale è l'upgrade naturale una volta implementato il guard canton-root-validity
- Suggested action: Implementare canton-root-validity guard in `infra/cloudflare-worker/locale-router.js` (verifica che il canton-root target risponda 200 prima di emettere 301), poi upgradare da soft-404 a 301 reale per (a) company-hub orphan con prefisso corretto ma hub pruned e (b) job-detail expired de/fr

### LINKED ISSUE #3054: follow-up(#3035): 1 item deferred — feat(ads): in-feed ad ogni 3 card tra gli annunci correlati (Closes #2960)
## Origine
- PR: #3035 feat(ads): in-feed ad ogni 3 card tra gli annunci correlati (merged 2026-06-29)
- URL: https://github.com/valerielinc-ops/frontaliere-si-o-no/pull/3035

### 1. Verificare allineamento push-slot ins-tag related-section nel twin SSG
- Source: adversarial check (reviewer bot, seconda review)
- Original text:
  > Allineamento push-slot dello static loader sulla pagina SSG. `adSlotHtml.ts:42-45` avverte che ADSENSE_LOADER_CONTENT fa una push per ogni ins-tag in ordine DOM e la push non e slot-bound. L'ins in-feed iniettato a meta related-list mantiene l'allineamento col trailing end-of-list multiplex? Il pattern rispecchia `jobCardListBody` (provato), ma non l'ho validato per la related-section in isolamento - un disallineamento renderebbe lo slot sbagliato.
- Funnel impact: monetizzazione (wrong ad slot served → RPM loss se il push si disallinea rispetto agli ins-tag del multiplex trailing)
- Rationale: l'AdSense loader in `adSlotHtml.ts:42-45` fa push per ogni ins-tag in ordine DOM senza slot-binding; la PR aggiunge ins-tag nella related-section SSG (`jobsSeoPagesPlugin.ts`) che non era presente nella versione provata di `jobCardListBody`. Un disallineamento silenzioso degraderebbe il CPM.
- Suggested action: leggere `adSlotHtml.ts:42-45` per capire se il push e posizionale o slot-bound; verificare tramite diff/codice se la posizione DOM degli ins-tag related coincide con lo slot-index atteso dal multiplex trailing; se disallineamento → aggiungere attributo `data-ad-slot` esplicito o ristrutturare la push.

### LINKED ISSUE #3080: follow-up(#3074): 1 item deferred — perf(llm): opt-in response cache + page-validation cache to cut free-tier consumption
## Origine
- PR: #3074 perf(llm): opt-in response cache + page-validation cache to cut free-tier consumption (merged 2026-06-29)
- URL: https://github.com/valerielinc-ops/frontaliere-si-o-no/pull/3074

### 1. Verifica: model identifier incluso nella chiave della cache persistente `data/jobs-ai-cache.json`
- Source: adversarial check
- Original text:
  > Interazione cache persistita x rotazione modello: sotto fallback, la cache stabilizza il verdetto al **primo** modello che ha risposto; un run successivo con il primario esausto avrebbe ri-classificato via fallback pre-PR. Esito deterministico -> nessuna regressione di de-index attesa, ma l'interazione cross-run non è stata osservata in diretta.
- Funnel impact: traffico (job indexing — un verdetto mis-classificato da un provider fallback si cristallizza nella cache persistente tra run → pagina job potenzialmente esclusa dall'indice in modo permanente)
- Rationale: `buildAiCacheKey` / `getCachedAiResponse` per `aiValidateJobDetailPage` potrebbe non includere il model identifier nella chiave; se un provider fallback classifica una pagina borderline diversamente dal primario, il verdetto resterebbe congelato tra run successivi. Il reviewer nota che l'esito è atteso deterministico ma "non è stato osservato in diretta" → needs triage.
- Suggested action: verificare se `buildAiCacheKey` in `scripts/lib/ai-models.mjs` include il model identifier nella chiave della cache persistente (`data/jobs-ai-cache.json`); se no, aggiungerlo per isolare verdetti cross-provider e prevenire cristallizzazione di verdetti da provider di qualità inferiore.

### LINKED ISSUE #3150: follow-up(#3144): 1 item deferred — fix(jobs): relocation bridge (not orphan tombstone) for live cross-canton legacy-TI URLs
Origine: PR #3144 (merged 2026-06-30) https://github.com/valerielinc-ops/frontaliere-si-o-no/pull/3144

Item 1 — Verifica: stash activeDriftRealPathByCompat keyed solo su compatPath — rischio canonical locale-sbagliato su prefix collision

Source: adversarial check (reviewer bot)

Original text: Wrong-locale canonical su key-collision — lo stash e keyed solo su compatPath (non su slug+locale). compatPath = prefix+slug e unico per (slug,locale) dato che prefix deriva da COMPAT_JOB_PATTERNS locale-specifici, ma non ho enumerato tutti i pattern per escludere due locali con lo stesso prefix. Se collidessero, lultimo .set vincerebbe e il canonical punterebbe alla pagina dellaltro locale.

Funnel impact: traffico / correttezza SEO (canonical locale-sbagliato su URL job cross-canton)

Rationale: Potenziale bug strutturale: se due locali condividono lo stesso prefix in COMPAT_JOB_PATTERNS, la Map keyed solo su compatPath verrebbe sovrascritta dallultimo .set, producendo bridge con canonical che punta alla pagina del locale sbagliato. Impatto diretto su circa 10k URL legacy-TI cross-canton in tutti i locali (FR/EN/DE/IT).

Suggested action: Enumerare tutti i COMPAT_JOB_PATTERNS locale-specifici in jobsSeoPagesPlugin.ts per verificare unicita del prefix per locale. Se puo collidere cambiare la Map key da compatPath a locale:compatPath (chiave composita) nello stash a L9940 e aggiornare il lookup self-healing a L12077 di conseguenza.

### LINKED ISSUE #3266: chore(ci): 46 non-crawler workflows share the same missing-GH_TOKEN gap as the crawler fix in #3265
## Contesto

PR #3265 ha fixato un gap strutturale nei 450 workflow `update-jobs-*.yml`: lo step "Load secrets from Remote Config" girava senza `GH_TOKEN`, causando fallimento silenzioso di ogni chiamata `gh` CLI interna allo script del job (es. shrink-guard/boilerplate-guard che tentano di aprire una issue `parser-broken`).

Grep dello stesso step-text esatto (`Load secrets from Remote Config` → `run: node scripts/load-rc-env.mjs`) mostra che altri **46 workflow non-crawler** (deploy/analytics/follow-up/ecc.) condividono lo stesso pattern e potenzialmente lo stesso gap, se uno qualsiasi degli script che invocano internamente `gh` gira in uno step successivo senza il proprio `GH_TOKEN` esplicito.

Non incluso in #3265 perché: classe di workflow diversa (non crawler), non implicata nei fallimenti crawler segnalati oggi — includerlo lì sarebbe stato scope creep non necessario a chiudere quel task.

## Da fare

- `grep -l "Load secrets from Remote Config" .github/workflows/*.yml | grep -v update-jobs-` per la lista dei 46 file.
- Per ciascuno, verificare se un qualunque step successivo nello stesso job invoca `gh` (issue/PR/comment) senza `env: GH_TOKEN` esplicito.
- Se sì: applicare la stessa fix (persist `GH_TOKEN` a `$GITHUB_ENV` nello step "Load secrets from Remote Config").
- Se nessuno dei 46 invoca `gh` internamente, chiudere come falso positivo (nessun impatto pratico anche se il gap esiste testualmente).

### LINKED ISSUE #3302: follow-up(#3300): 1 item deferred — fix(seo): extend cluster-301 candidate map to en/de/fr
## Origine
- PR: #3300 fix(seo): extend cluster-301 candidate map to en/de/fr, verify full map coverage (merged 2026-07-02)
- URL: https://github.com/valerielinc-ops/frontaliere-si-o-no/pull/3300

### 1. `services/searchQueryBoilerplate.mjs`: `stripSearchQueryBoilerplate` behavior delta for single-boilerplate-word bodies may produce wrong redirect targets on future map regeneration
- Source: adversarial check
- Original text:
> `services/searchQueryBoilerplate.mjs:568-585`: the new `stripSearchQueryBoilerplate` never empties a body that IS exactly one boilerplate word (e.g. `"lavoro"`) — old `LEADING_BOILERPLATE` logic explicitly set `body = ''` in that case. Verified against the actual committed map that this never currently produces a wrong "specific" target (0 hits scanning for `/ricerca-lavoro/`-shaped targets), and the docblock's self-heal-on-regen property bounds the blast radius, but it is a real behavior delta from the old code that is not called out in the PR body.
- Funnel impact: traffico (redirect quality for search cluster legacy URLs → SEO)
- Rationale: behavior delta from old code; future map regeneration could produce wrong "specific" targets for single-word query bodies (e.g. input `"lavoro"` produces non-empty body instead of `''` → wrong cluster target URL), silently degrading 301 redirect quality for those cluster paths

### LINKED ISSUE #3303: follow-up(#3299): 1 item deferred — fix(seo): re-tighten spa-locale title-length cap + migrate fuel-slug lists to shared module
## Origine
- PR: #3299 fix(seo): re-tighten spa-locale title-length cap + migrate fuel-slug lists to shared module (merged 2026-07-02)
- URL: https://github.com/valerielinc-ops/frontaliere-si-o-no/pull/3299

### 1. `scripts/audit-dist-multi.mjs:L765,L1060`: fuel-slug classifiers still use pre-migration inline alternation, missing 5 legacy-alias slugs
- Source: reviewer nit
- Original text:
> `scripts/audit-dist-multi.mjs:L765,L1060`: Nit — `sdClassifyPage()`/`sdValidateFuelMerchantProduct()` still use the pre-migration inline fuel-slug alternation (`prezzi-(diesel|benzina)|diesel-price-switzerland|...`), missing the 5 legacy-alias slugs `FUEL_SECTION_RX` now covers. Same drift class this PR fixes in `build-legacy-aliases.mjs`/`report-dist-bytes-by-plugin.mjs`, and this exact file is named in the sticky sibling-check comment, but the PR body only declares `validate-structured-data-completeness.mjs` out of scope — it does not mention that `audit-dist-multi.mjs` carries a "verbatim" copy of that same excluded logic (comment at L752). If the same "can't safely validate a structural change pre-deploy" reasoning applies here too, say so in `## Non implementato`; otherwise wire it to `isFuelSectionPath()` like the other 3 audit classifiers. Funnel-critical (fuel Product JSON-LD completeness gate), actionable — normal follow-up candidate.
- Funnel impact: funnel (fuel Product JSON-LD completeness gate — structured data validation for fuel pages)
- Rationale: fuel Product JSON-LD validation gate silently misses pages at 5 legacy-alias slug paths; reviewer explicitly flagged as follow-up candidate; same drift class as what PR #3299 fixed in the other two scripts

### LINKED ISSUE #3304: follow-up(#3292): 1 item deferred — fix(seo): harden sitemap-news slug-sync guard for swiss articles
## Origine
- PR: #3292 fix(seo): harden sitemap-news slug-sync guard for swiss articles (merged 2026-07-02)
- URL: https://github.com/valerielinc-ops/frontaliere-si-o-no/pull/3292

### 1. `tests/sitemap-slug-integrity.test.ts`: slug-sync gate verifies IT canonical `<loc>` only — stale/wrong EN/DE/FR hreflang alternates pass silently
- Source: adversarial check
- Original text:
> The gate only checks the IT canonical `<loc>`; a stale/wrong EN/DE/FR `hreflang` alternate (`schweiz-artikel`/`articles-suisse`/`swiss-articles` paths) would not fail this test. Same limitation exists identically in the blog block being mirrored (pre-existing design, not a regression introduced here).
- Funnel impact: traffico (stale/wrong hreflang alternates in sitemap-news.xml affect international SEO routing and may trigger Google hreflang validation errors)
- Rationale: gate extension in PR #3292 covers IT canonical `<loc>` only; wrong EN/DE/FR hreflang alternates for swiss articles would pass the gate silently, enabling the same class of bug as #3120 (3 dead hreflang URLs) to recur undetected on the alternate-locale paths

### LINKED ISSUE #3310: en/de/fr Ticino company-hub 404: solothurner-spitaler-soh missing from per-locale hub emission
## Origine
- Found by: agent working #3173 (sitemap 404-risk served-set fix), PR pending in bundle round 3
- Related: #3173 (fixed the served-set audit-script gap; this is a separate, real bug found via curl)

### 1. Company hub page missing for en/de/fr, present for IT
- Live curl confirms: `cerca-lavoro-ticino/azienda-solothurner-spitaler-soh/` → 200 (IT). `en/find-jobs-ticino/company-solothurner-spitaler-soh`, `de/jobs-im-tessin/unternehmen-solothurner-spitaler-soh`, `fr/trouver-emploi-tessin/entreprise-solothurner-spitaler-soh` → real 404 (not a served-set audit artifact — actually 404 live).
- Root-cause hypothesis (not fully verified, flagged as hypothesis not fact): `canonicalCompanySlugBuild` produces identical slugs cross-locale (only special-cases "lidl"), so not a slug-mismatch. Looks like a per-locale `validJobs` filtering gap in `build-plugins/jobsSeoPagesPlugin.ts` — likely this company's job(s) lack an en/de/fr translated description, dropping them from the per-locale hub-emission loop while IT still emits (IT doesn't require translation).
- Funnel impact: traffico (3 dead hub URLs currently, but the pattern — any company with IT-only description — would recur for other companies as data changes)
- Next step: confirm the missing-translation hypothesis by inspecting this company's job record fields, then either backfill the translation or make the per-locale hub loop fall back gracefully (e.g. don't gate hub emission on translated description, only on job existence) instead of silently dropping the hub page.

valerielinc-ops:
## Implementato

- Closes #3080 — `aiValidateJobDetailPage`'s persistent AI-verdict cache key now folds in the actually-served model identifier (via new `getPreferredModel()` peek in `ai-models.mjs`), so a fallback-tier verdict on a borderline job page can no longer get frozen into `data/jobs-ai-cache.json` under the primary model's identity across runs. Sibling fix: `callLLM`'s opt-in `_responseCache` had the same pre-call-vs-served-model key mismatch, fixed alongside. Sibling audit of 5 other AI-cache call sites: 2 are false positives (deterministic non-AI sentinel values share the key, widening would incorrectly gate on model availability), 3 already have independent quality guards busting bad cache regardless of source model.
- Addresses item 1 (of 3), item 2 (of 3), item 3 (of 3) of #3216 — all 3 items done, orchestrator will manually **shut** #3216 post-merge citing this commit. Item 1: found+fixed a THIRD sibling of the version-skew pattern-drift class (root `index.html`'s hand-maintained inline `<script>`, which cannot import `resilientImport.ts`, was missing the `'Importing a module script failed'` substring check already present in the other 5 consumers) — now parameterized `constants.ts`'s `SELF_HEAL_SCRIPT_CONTENT` against the same `CALL_TIME_SKEW_SOURCE`/`MODULE_LINK_SKEW_SOURCE`/`CHUNK_LOAD_ERROR_SOURCE` single-source patterns (manually reconciled during cherry-pick against an earlier round's independent fixes in the same script — buffer-size raise, unconditional DOM-scan union, `credentials:'same-origin'` — all preserved). Item 2: exhaustive double-inclusion audit across every `EARLY_BOOT_SCRIPT` consumer, no double-inclusion found. Item 3: documented 3 previously-undocumented live Cloudflare Cache Rules in `infra/cloudflare/rules.md` with rollback commands. 81 targeted tests green (index-html-self-heal, earlyBootSelfHeal, resilient-import, ChunkLoadErrorBoundary, shared-components-smoke, stable-asset-names). Addendum: a second, independent investigation of #3216 found one more gap in the same file — `index.html`'s inline error handler used a literal leading space (`/ is not a function\b/`) instead of the canonical `\b` word boundary (`/\bis not a function\b/`) used in `resilientImport.ts`, on all 3 CALL_TIME_SKEW patterns. Fixed for byte-parity; 14/14 targeted tests green.
- Addresses item 2 (of 2) of #3104 — replaced the broken `gh run download` + `tar` pattern (fails on the >1GB Pages artifact) with the proven `gh api .../artifacts/<id>/zip` flow across all 4 sibling workflows (`seed-title-baselines.yml`, `seed-bfs-depth-baseline.yml`, `seed-orphan-pages-baseline.yml`, `measure-deploy-delta.yml` — the last kept inline due to its narrow sparse-checkout, others share a new `.github/actions/fetch-pages-artifact` composite action, also applied to the already-merged `seed-text-html-ratio-baseline.yml` per sibling-fix discipline). `actionlint` clean. Item 1 (live baseline reseed) intentionally untouched — see Non implementato.
- Addresses item 1 (of 3), item 2 (of 3), item 3 (of 3) of #3209 — all 3 items done, orchestrator will manually **shut** #3209 post-merge citing these commits. Item 1: extracted duplicated slug-derivation logic into exported `deriveAndSanitizeArticleSlugs()` (`scripts/create-article.mjs`), `registerArticleFiles()` now calls it and returns the finalized slugs; removed `publish-journalist-article.mjs`'s drift-prone copy (manually reconciled during cherry-pick to keep an earlier round's #3010 EN/DE/FR `checkTranslatedSlugCollisions` guard working against the new single-source derivation, called explicitly before `registerArticleFiles()`). Follow-up consolidation commit (sibling-pattern-fix, AGENTS.md §6): `validate()` still hand-rolled the identical slug-format regex chain 3× inline (id synthesis + IT-locale fallback + en/de/fr fallback loop) — now calls the existing `slugifySlugPart()`; and `publish-journalist-article.mjs` still hand-rolled its own final-URL builder (`buildPublishedUrls()`/`HUB_SLUG`), which had drifted and **omitted the `/en`/`/de`/`/fr` locale prefix** router.ts's `buildPath()` actually uses — producing wrong links in the "your article is live" email/Firestore field. New `buildArticlePublishedUrls()` (single source of truth, same prefix convention as `buildSectionSitemapUrls()`'s hreflang links) fixes this; `registerArticleFiles()` now returns `{ slugs, publishedUrls }`. Item 2: `check-journalist-article-links.mjs` now skips docs published <5min ago (5-min window inside the 15-min cron cadence) to avoid guaranteed-404 on a just-published article's own URL. Item 3: `resolveHeroImage()`'s timeout was already fixed by merged PR #3210; added missing test coverage. 79 targeted tests green (article-slug-derivation ×10 incl. 4 new regression tests, check-journalist-article-links-ordering, publish-journalist-hero-image-timeout, events-digest-article-generator, create-article-gates, + 20-file wider article/journalist sweep).
- Closes #3010 — canonicalizes the 3 shadowed svizzera near-duplicate article pairs (owner sign-off obtained 2026-07-02; chose the non-destructive canonical-tag option over removal, consistent with the repo's firm anti-page-deletion rule). Pair designations (shadowed → authoritative) taken directly from PR #3000's own body, which already documented which slug was unreachable due to the `REVERSE_SWISS` last-wins bug: `tassazione-frontalieri-2026-nuovo-accordo` → `frontaliere-tassazione-2026-regole-accordo`; `tassazione-frontalieri-oltre-20km-confine` → `tassazione-frontalieri-oltre-20km`; `educatore-infanzia-ticino-lavoro` → `lavorare-educatore-infanzia-ticino`. Implemented in `build-plugins/ogPagesPlugin.ts` (the static SSG emitter, not the runtime `seoService.ts`) via a new per-slug override map (`data/swiss-article-canonical-overrides.json`, 12 entries across 4 locales), mirroring the repo's existing `jobsSeoPagesPlugin.ts`/`job-canonical-overrides.json` cannibalization-fix convention per sibling-fix discipline. Shadowed pages stay in sitemap/RSS/routing — canonical-hint only, no removal/noindex. 37 targeted tests green (swiss-article-canonical-overrides, job-detail-seed, criticalCssRootHeight); broader regression spot-check (seo-completeness 44431 cases + 6 more ogPagesPlugin consumers) also green per the agent's report.
- Addresses item 1 (of 2), item 2 (of 2) of #2861 — both items done, orchestrator will manually **shut** #2861 post-merge citing this commit plus the earlier item-2 commit already on this branch. Item 1: migrated 9 more workflows (`auto-merge-sweep.yml`, `pr-collision-detector.yml`, `discover-404s.yml`, `discover-404s-via-cloudflare.yml`, `monitor-gsc-seo.yml`, `recycle-stale-prs.yml`, `generate-article.yml`, `traffic-data-freshness.yml`, `sync-gsc-orphans.yml`) to the GitHub App token-minter pattern (mirrors `pr-autorebase.yml`'s reference implementation), widened `GH_TOKEN`/fallback-guard expressions so no scope silently drops if App minting fails. 3 workflows left as documented PAT exceptions (`gh-pat-expiry-monitor.yml` monitors the PAT itself; `pr-redflag-fixer.yml`'s `workflow`-scope requirement on the App installation is unverified, not fabricated; `snapshot-jobs-weekly.yml`'s `GITHUB_PAT` reference is an unrelated LLM-provider bearer key, not git/gh auth). `actionlint` identical before/after (224/224, zero new findings). No workflows dispatched live (owner authorization covers the code migration; live validation happens on each workflow's next natural trigger post-merge).
- Addresses item 3 (of 5) of #3036 — combined with item 4 (already in this PR) and items 1/2/5 (already resolved by prior merged PRs), all 5 items of #3036 are now complete — orchestrator will manually **shut** #3036 post-merge citing this commit plus the prior ones. Item 3: mirrors tio-agenda flyer images to this site's own CDN storage (reusing the existing `mirrorEventImage` mechanism already used by the guidle/myswitzerland crawlers — tio-agenda was the only one still hotlinking) and wires a full `ImageObject` into event JSON-LD via the sitewide `imageObjectLd()` builder. Found and fixed an active regression in the process: `eventsSeoPagesPlugin.ts` was unconditionally embedding the raw third-party `biglietteria.ch` flyer URL into production JSON-LD. License fields (`license`/`acquireLicensePage`/`creator`) are left unset since no per-image license text exists at the source — falls back to the site's own terms page rather than fabricate a claim; `creditText` is set to the honestly-known `sourceName`. Manually reconciled during cherry-pick against item 4's independent `warnIfLowConfidenceComuneShare` addition in the same file (both kept, main()'s control flow reordered so the warning still fires before the now-earlier empty-events guard, then mirroring runs before both the dry-run preview and the write). 110 targeted tests green (events-pipeline, events-detail-pages, events-nationwide-sources, imageObjectLd). Post-merge: a live `crawl-tio-agenda.mjs` run is needed to backfill mirrored images for the 136 existing events (graceful `SITE_IMAGE` fallback until then, no breakage).
- Addresses item 1 (of 3) of #3206 — verified directly against production (Firebase Admin SA): both one-time scripts this item asked to `--apply` (`scripts/dev/backfill-newsletter-subscriber-auth.mjs`, `scripts/dev/reactivate-false-positive-bounces.mjs`) were **already run in production on 2026-07-01 ~09:11 UTC**, before this follow-up issue existed (529 Auth accounts backfilled vs ~522 estimated; 2017 bounce docs reactivated vs ~1867 estimated; fresh dry-run today confirms 0 remaining candidates on both). No code gate was missing — both scripts already defaulted to dry-run with explicit `--apply` opt-in, the established repo-wide convention (confirmed via sibling-pattern sweep, ~20+ scripts use the identical convention, no gap). Added a docblock status note to both scripts recording the completed run so the issue's stale text doesn't mislead a future reader into re-running. Combined with item 2/3 (done in earlier rounds), all 3 items of #3206 are now genuinely complete — orchestrator will manually **shut** #3206 post-merge citing this commit plus the earlier item-2/3 commits.
- Addresses item 1 (of 2), item 2 (of 2) of #3135 — both items done, orchestrator will manually **shut** #3135 post-merge. Item 1 (this commit): removed the `maileroo` manual force-select path — dropped from `send-newsletter.yml`'s `workflow_dispatch.provider` choice and from `SINGLE_PROVIDERS` in `scripts/send-newsletter.mjs`, so an operator can no longer force-route a send through Maileroo bypassing normal cascade rotation. Left untouched: `sendViaMaileroo()`/`PROVIDERS` entry in `scripts/lib/email-cascade.mjs`, the automatic cascade-rotation tier (separate mechanism, DKIM/DMARC-verified and re-enabled by already-merged PR #3154 — note the original issue's "not fully authenticated" premise for Maileroo is now outdated, this commit is independently defensible as removing an operational cascade-bypass override). Item 2: already fixed by merged PR #3136 (verified against current `dmarc-monitor.mjs`, not redone). 47 targeted tests green (newsletter-provider-force-select (new), dmarc-monitor, email-cascade-burst).
- Sibling gap fix (own commit, no linked issue): the #3206-item3 Firestore-transaction wrap for `maybeEscalateSoftBounce` updated mailjet/maileroo/resend webhook-core test mocks with a matching `docRef.firestore.runTransaction` shim, but missed the identical mock shape in `tests/newsletter-mailgun-webhook-core.test.ts` / `tests/newsletter-mailtrap-webhook-core.test.ts` — both failing CI with "Cannot read properties of undefined (reading 'runTransaction')". Applied the same shim to both, all 5 ESP webhook-core suites now green (59/59).
- **#2761** — validate-dist tar-pack: verify byte-identity vs git-clone push path + gate rehydrate fast-path on tar-completeness sentinel check (was silently passing on truncated/partial extraction).
- **#3054** — audited AdSense in-feed push-slot mechanism for the new related-jobs `<ins>` tags: confirmed slot binding is attribute-based (`data-ad-slot`) not push-order-based, so no misalignment is possible; added regression test codifying the invariant.
- **#3036** (item 4 only) — `resolveComune`/`resolveComuneNationwide`: extracted shared `matchComuneNames()`, added `warnIfAmbiguousComuneMatch()` + `warnIfLowConfidenceComuneShare()` observability for uncertain/region-fallback matches.
- **#3047** (both items) — `schedule-fb-events-daily.mjs`: pre-flight HEAD check (bounded timeout) before posting a comune landing-page link to Facebook; skips post on 4xx/timeout, fail-open on ambiguous 5xx. Item 1 (crawler expansion feasibility) investigated and documented as infeasible with existing data — see Refs below.
- **#2918** (all 3 items now done — item 1 already covered by merged #3300; orchestrator will manually shut #2918 post-merge citing these commits) — `build-search-cluster-301-map.mjs`: fail loud (non-zero exit) on unexpected/empty `indexed-cluster-urls.json` shape instead of silently producing an empty map, plus a second independent guard (`assertNonEmptyLegacyMap`) for a validly-shaped file whose URLs match no locale pattern. Item 3 (periodic regen + prune): new `scripts/prune-search-cluster-301-map.mjs` drops national-cluster-page 301 targets that are no longer live, modeled on `prune-404-compat-paths.ts`'s pattern; exports the shared helpers so it and tests can import `build-search-cluster-301-map.mjs` without triggering its live network fetch. Cron/schedule wiring (the other half of item 3, dispatched as a separate follow-up worktree task once the regen/prune scripts existed): new `.github/workflows/refresh-search-cluster-301-map.yml` runs `build-search-cluster-301-map.mjs` then the prune pass weekly (Mon 06:00 UTC, 50min after the sibling `refresh-indexed-cluster-urls.yml` this consumes), validates the committed-snapshot invariant (`tests/search-console-compat.test.ts`'s full-coverage check) before committing, then commits+pushes via the standard `git-push-with-retry.sh --regenerate-cmd` pattern — no secrets needed, both scripts only hit the public site's own sitemaps. Found and fixed a real bug en route: `build-search-cluster-301-map.mjs`'s previously-unconditional `main()` call fired an unawaited live-network regeneration as a side effect of import (the prune script imports it for its exported helpers) — guarded behind an `invokedDirectly` check, zero behavior change to the direct-CLI path. `actionlint` clean; 119 targeted tests green (prune-search-cluster-301-map ×7, search-console-compat ×20 incl. the 9446-entry full-coverage check, related-search-clusters ×85, locale-router-legacy-search-cluster-301 ×7).
- **#3150** — documented + regression-guarded the compat-prefix locale-uniqueness invariant in `jobsSeoPagesPlugin.ts` (enumeration proof no two locale compat prefixes can collide).
- **#2764** — `relatedSearchJunkTerms.mjs`: stop dropping legitimate "stipendio `<role>`" salary-intent keyword candidates (was wholesale-dropping ~4.8k live IT candidate slugs).
- **#2861** (item 2 only) — new `app-auth-health-monitor.yml`: daily scheduled check that GitHub App token minting still works, alerts via the existing issue-creator pattern (mirrors `gh-pat-expiry-monitor.yml`).
- **#2886** — new R2 CDN janitor (`_janitor_cdn_r2` in `deploy-it-pages-prep.sh`): lists stale `assets/` objects vs the active build set + grace period; **dry-run only by default**, real deletion requires both an explicit function arg AND an explicit env opt-in, neither wired into any automatic call site.
- **#3206** (item 2 only) — applied the shared bounce-severity classification (from #3189) to `job_alert_subscribers` across all 5 ESP webhook cores (Mailjet/Mailgun/Maileroo/Mailtrap/Resend) — soft bounces no longer hard-flip subscribers to `bounced`.
- **#3106** — proved undici `Agent` `headersTimeout`/`bodyTimeout` are actually enforced end-to-end at runtime (new test spins a real hanging `http.Server`, asserts rejection near the ~500ms budget, not just constructor shape); promoted `undici` from transitive to direct dependency, closing a silent-degradation gap where a failed `import('undici')` would be swallowed. Added a second, more surgical test isolating the item-2 claim directly (raw global `fetch` + raw `undici.Agent` dispatcher against a hung TCP server, no `AbortSignal`, no local-LLM plumbing in between) — confirms the dispatcher itself is honored by this runtime's `fetch`, not just that the composed retry/timeout stack happens to abort in time.
- **#3104** (item 2 only) — applied the proven gh-api-zip dist-download fix to the 4 sibling seed/measure workflows (`seed-title-baselines.yml`, `seed-bfs-depth-baseline.yml`, `seed-orphan-pages-baseline.yml`); `measure-deploy-delta.yml` is a **false positive** (diffs artifact tarballs by content hash, never materializes `dist/` — documented in-file, left unchanged).
- **#3015** — `infra/cloudflare-worker/locale-router.js`: canton-root-validity guard (verifies canton-root target responds 200 before emitting a 301), upgrading company-hub-pruned-orphan and job-detail-expired-de/fr from soft-404 to real 301.
- **#2772** (all 4 items) — `assemble-jobs-dataset.mjs`: new `acceptInferredCantonForFill()` guard rejects canton-fill inferences outside `TARGET_CANTONS` (item 3); also confirmed `TARGET_CANTONS` already covers all 26 cantons since an earlier CH-wide expansion, so the originally-described bug is currently unreachable — guard kept as forward-looking protection. Items 1/2/4: 48 re-drifted `data/all-known-job-slugs.json` canton-section entries reconciled as a one-time audit per the issue's scope (item 1); new `cantonEmptyNoSignal` counter+sample tracking surfaces previously-silent no-signal jobs (item 2); legacy-pin-overwrite already resolved by merged #3164, independently re-confirmed (item 4).
- **#2935** — follow-up of #2931: confirmed `JOBLIST_AD_MAX_PER_LIST=12` cap behaves correctly across the SPA main list's mobile infinite scroll (`JobBoard.tsx` `loadMoreMobile`) — `shouldPlaceInfeedAd` sees the absolute cumulative card position each render, so ads plateau at 12 past 36 accumulated cards with no re-shuffling. Added regression tests pinning the cadence+cap contract; cap value kept at 12 (policy-safety default, unchanged pending explicit owner override).
- **#3304** — extended `tests/sitemap-slug-integrity.test.ts`'s hreflang gate (previously IT-canonical-`<loc>`-only) to bidirectionally validate every EN/DE/FR `<xhtml:link>` alternate, for both the pre-existing Blog block and the newer Swiss-article block (same gap in both — confirmed same-file sibling, not a regression).
- **#3302** — `build-search-cluster-301-map.mjs`'s `nationalSlugBody()`: an all-boilerplate-token body (e.g. legacy `/ricerca-lavoro/`, or a nation-suffix-only leftover) now resolves to `null` instead of being treated as a non-empty "specific" cluster body; scoped to this one caller, the shared `stripSearchQueryBoilerplate()` is untouched (deliberately never empties — pinned by an existing display-facing test).
- **#3303** — `audit-dist-multi.mjs`'s `sdClassifyPage()`/`sdValidateFuelMerchantProduct()` migrated from a stale inline fuel-slug alternation to the shared `FUEL_SECTION_RX`, picking up the 5 legacy-alias slugs for fuel Product JSON-LD validation coverage.
- **#3119** (item 3 only; items 1/2 already resolved by merged #3129/#3134) — `aldi-suisse-job-parser.mjs`'s `parseAldiSearchResults`: added dedup by the REST feed's stable `rmk_id` alongside existing URL-based dedup, so two rows sharing a physical job with different `sys_language_uid` (and thus different URLs) collapse to one result instead of publishing duplicate-language variants as separate listings.
- **#3149** (both items) — raised `performance.setResourceTimingBufferSize(1000)` as the first statement in all three chunk-skew self-heal bootstrap copies (`services/resilientImport.ts`, `index.html` inline script, `build-plugins/constants.ts`'s `SELF_HEAL_SCRIPT_CONTENT`) so a long/content-heavy session doesn't evict the skewed chunk's Resource Timing entry before recovery inspects it, plus a defensive call inside `bustAssetHttpCache()` itself; also unioned the DOM `<script>`/`<link>` scan into the asset-URL set instead of only using it as an empty-buffer fallback, so statically-referenced assets stay covered even under partial eviction (dynamically-`import()`'d chunks remain a documented structural gap — they never create a DOM node, buffer-size raise is the real mitigation there); and fixed a cache-key mismatch — the cache-busting `fetch()` calls used `credentials:'omit'`, but a module script with no `crossorigin` attribute (this build's emit shape) defaults its whole module graph to `'same-origin'` credentials mode, so `'omit'` only coincidentally matched behavior when `ASSET_CDN` was set (cross-origin) and diverged from real browser behavior in the same-origin (`ASSET_CDN` unset) deploy shape — changed to `credentials:'same-origin'`.
- **#3078** — verified the sessionStorage reload-ceiling unification item is already resolved (PR #3077/#3184/#3233 unified onto a single `_swReloadCount` key); no code change needed for item 1.
- **#3089** (all 3 items) — verified all 3 adversarial-check items were already fixed upstream (gated-slice write order was always safe; `companyKey` mismatch warn fixed by #3122; catch-all silent-swallow fixed by #3107) but had zero test coverage; added a 74-case regression test scanning every `runDedicatedBaseCrawler` call site for the gated-write-order invariant, plus unit tests locking in the other two fixes' behavior.

All fixes were produced independently by parallel worktree-isolated agents (one fix-only commit each, no PR of their own) and bundled here into a single PR to cut down `pr-review-loop` cycles.

Closes #3054
Closes #3150
Closes #2764
Closes #2886
Closes #3015
Closes #3304
Closes #3302
Closes #3303

Addresses item 1+2 (all items) of #2761 — both items fully implemented in this PR (byte-identity check + tar-completeness gate). Issue #2761 is a 2-item aggregate; `Closes` keyword on a multi-item aggregate is structurally blocked by `pr-body-contract.yml`'s anti-recidivism gate regardless of actual completion, so the orchestrator will manually shut issue #2761 post-merge with a commit citation instead of relying on this keyword.
All 3 items of #2918 now done (item 1 via merged #3300, items 2+3 in this PR, including item 3's cron-wiring workflow) — orchestrator will manually shut issue #2918 post-merge citing these commits. Live-fire validation of the new weekly cron (`refresh-search-cluster-301-map.yml`) is a post-merge observation, not a merge blocker — the workflow's own commit step is gated behind the existing committed-snapshot test.
Addresses item 1+2 (all items) of #3106 — both adversarial-check items (undici Agent timeout support in the installed version; Node global-fetch dispatcher honoring on the actual runtime) are covered by the single end-to-end enforcement test. Issue #3106 is a 2-item aggregate; same `Closes`-blocked-on-aggregate restriction applies — orchestrator will manually shut issue #3106 post-merge with a commit citation.
All 5 items of #3036 now done (item 4 in this PR, item 3 in this PR, items 1/2/5 already resolved by prior merged PRs) — orchestrator will manually shut issue #3036 post-merge with a commit citation.
Addresses item 1+2 (all items) of #3047 — item 2 (pre-flight HEAD landing check before FB post) already fixed by a prior merged commit; item 1 (crawler expansion feasibility to individual comune sites/other cantons) investigated and resolved as genuinely infeasible with existing data (no comune→URL map exists in any dataset; "altri cantoni" half already moot per #3125's nationwide aggregator sources), documented in `scripts/lib/events-utils.mjs` so it isn't re-researched. Issue #3047 is a 2-item aggregate; orchestrator will manually shut #3047 post-merge with a commit citation.
Both items of #2861 now done (item 2 already on this branch, item 1 in this PR) — orchestrator will manually shut issue #2861 post-merge with a commit citation.
All 3 items of #3206 now done (items 2/3 in earlier rounds, item 1 in this PR) — orchestrator will manually shut issue #3206 post-merge with a commit citation.
Both items of #3135 now done (item 2 already fixed by merged PR #3136, item 1 in this PR) — orchestrator will manually shut issue #3135 post-merge with a commit citation.
Refs #3104 (item 2 done; item 1 — reseed `data/text-html-ratio-baseline.json` via live `seed-text-html-ratio-baseline.yml` dispatch with an `eventi`-inclusive `deploy_run_id` — requires a live workflow dispatch + committing the regenerated baseline, not a code change; tracked as its own follow-up)
All 4 items of #2772 now done (item 3 already in this PR; items 1/2/4 reconciled — 48 re-drifted `all-known-job-slugs` entries fixed, no-signal self-heal residual now tracked/logged, legacy-pin-overwrite already resolved by merged #3164) — orchestrator will manually shut issue #2772 post-merge with a commit citation.
Refs #3119 (item 3 done here; items 1/2 already resolved by merged #3129/#3134 — all 3 items now complete across the three PRs; orchestrator will manually shut issue #3119 post-merge citing all three commits, since a bare `Closes` on this multi-item aggregate title is blocked by `pr-body-contract.yml` regardless)
Addresses item 1+2 (all items) of #3149 — both adversarial-check items (Resource Timing buffer eviction risk; credentials-mode cache-key mismatch) fixed in this PR. Issue #3149 is a 2-item aggregate; same `Closes`-blocked-on-aggregate restriction applies — orchestrator will manually shut issue #3149 post-merge with a commit citation.
Refs #3078 (item 1 verified already resolved, no code change; item 2 — bundle-size-delta OOM measurement — remains blocked, unrelated to this PR's scope)
Addresses item 1+2+3 (all items) of #3089 — all 3 already fixed upstream (#3122, #3107) or already-safe, now locked in with regression tests here. Issue #3089's "3 item deferred" title is a multi-item-aggregate pattern; same `Closes`-blocked-on-aggregate restriction applies — orchestrator will manually shut issue #3089 post-merge with a commit citation.
Addresses generator-hardening half of #3010's item 1 — translated-slug collision guard, previously IT-only inline in `checkForDuplicates()`, extracted into standalone `checkTranslatedSlugCollisions()` reused by both generators (`create-article.mjs` AI path and `publish-journalist-article.mjs`, which had zero cross-locale check before). Content-canonicalization half (the 3 near-duplicate article pairs) stays owner-gated per anti-cut rule — #3010 remains open until owner decides canonical/remove.
Addresses item 1+3 (all items) of #3055 — UBS Taleo siteid loop now covers apprenticeship (5054) + graduate (5131) alongside existing 5012; UBS region-guard bypass fixed (populated non-Swiss region can no longer be overridden by a city-name-alias match — pinned with a real-world Baden AG vs. Baden bei Wien, Austria fixture); Galenica canton-collision hardened (`state` must be blank before falling back to city-alias inference). Issue #3055 has multiple numbered items in its body (aggregate pattern); `pr-body-contract.yml`'s gate blocks `Closes` on it regardless of title wording — orchestrator will manually shut #3055 post-merge with a commit citation.
Addresses item 1+2 (all items) of #2852 — correction to an earlier round's entry: the issue is a 2-item aggregate, not 3 (re-verified against the issue's own body). Item 2 (instant webhook reactivation / win-back routing verification) is done across two commits: (a) traced the win-back resubscribe click definitively to the SPA path (`App.tsx` → `captureNewsletterSubscriber`), not the CF handler; confirmed `resubscribed_at` was never stamped there, making `subscriberSunset.mjs`'s re-consent guard dead code in production — fixed with an unconditional `resubscribed_at`/`resubscribedAt` stamp gated on `sourceChannel === 'resubscribe_link'`; (b) a subscriber the weekly `newsletter-sunset.mjs` cron already marked `inactive` was only reactivated on the next weekly run, not on the open/click that proves re-engagement — added a shared `functions/src/lib/subscriberReactivation.js` helper, wired into all 5 ESP webhook cores (Mailgun/Maileroo/Mailjet/Mailtrap/Resend) per sibling-pattern-fix discipline, with Resend needing the reactivation applied after `buildSubscriberUpdate()`'s pending-promotion side effect so an `inactive` subscriber doesn't land at `confirmed` without the reactivation fields. Item 1 (job-alert sunset v2 classifier, separate data model from the newsletter) is now also done: verified the stated prerequisite (open/click tracking analogous to `open_count`/`click_count`) is satisfied across all 5 providers' job-alert branches, then added `scripts/lib/jobAlertSunset.mjs` (thresholds `JOB_ALERT_SUNSET_MIN_DELIVERED=12`/`JOB_ALERT_SUNSET_MIN_AGE_DAYS=120`, `delivered_count` used as the ignored-sends proxy since no reliable send-time counter exists), `scripts/job-alert-sunset.mjs` runner, a weekly `.github/workflows/job-alert-sunset.yml` cron, and `isJobAlertExcluded()` wired into `send-job-alerts.mjs`'s existing suppression check — without touching hard-suppression. Deliberately out of this scope: a graduated win-back email stage for job alerts (goes straight candidate→sunset instead; would need new 4-locale email copy, not requested by the issue text). Fixed a sibling test-mock gap found during cherry-pick: the new instant-reactivation tests' Resend mock was missing the `docRef.firestore.runTransaction` shim required since #3206 item 3 wrapped Resend's status read/write in a transaction — added, matching the shim already used in `tests/newsletter-resend-webhook-core.test.ts`. 115/115 targeted tests green (newsletter-webhook-instant-reactivation, job-alert-sunset, email-suppression, subscriber-sunset, all 5 ESP webhook-core suites). #2852's aggregate title blocks `Closes`; orchestrator will manually shut #2852 post-merge citing these commits.

- **#3266** (closed as false positive, no code change) — audited all 46 non-crawler workflows sharing the `Load secrets from Remote Config` step-without-`GH_TOKEN` pattern from #3265; confirmed 0 real gaps (either nothing downstream shells out to `gh`, or the calling step already has its own explicit `GH_TOKEN`). Closed #3266 directly per its own stated resolution path ("chiudere come falso positivo" if no real gap found) — no `Closes` keyword needed here since it's already closed. Adjacent finding surfaced during that audit, NOT part of #3266's own scope: `publisher-jobs-sync.yml`'s "Build publisher-submitted slice" step calls `build-publisher-jobs.mjs` → `writeJobsCrawlerSlice()`, which contains the same `gh issue create` guard call the audit found dead-code-unreachable in `send-job-alerts.yml`/`send-newsletter.yml` — but here it's live, and the step runs before this workflow's own `Load secrets from Remote Config` step even exists in the job order (that step only loads a deploy-trigger PAT further downstream, conditioned on `has_changes`). Different bug class (step ordering vs. missing persist) but same silent-`gh`-failure consequence — fixed by giving the slice-build step its own explicit `GH_TOKEN`, matching the pattern this same workflow already uses for its "Report failure" step. `actionlint` clean.

- **#3203** (workflow-failure, root cause fixed) — the "Evergreen Article Refresh Audit" scheduled workflow crashed on its last 3 monthly runs (2026-05-01/06-01/07-01) with `SyntaxError: Invalid regular expression: missing /`. Root cause: `scripts/audit-evergreen-articles.mjs`'s `parseArticles()` stripped the `Article` TS interface with a regex that stops at the FIRST `}` — a JSDoc comment inside that interface mentions the route pattern `` `/autori/{authorSlug}/` ``, whose own balanced braces closed the match early and left a stray `/` from the comment feeding `new Function`, producing the crash. Not malformed source data — a bug in the script's own naive regex-based TS stripping. Fixed with brace-depth counting instead of a first-`}` regex; also closed two more latent breaks in the same eval path found while fixing (`satisfies Article[]` wasn't stripped; the deliberately-dropped `cdnBlogImage` import had no stub, so calling it during eval would throw). Sibling check: other `new Function(`-based TS-strip/JSON-eval sites in `scripts/` strip a narrower construct or parse unrelated scraped HTML, and are already try/catch-guarded — ruled out as a different, safer construct, not the same bug class. Re-verified on this bundle's HEAD (not just the agent's own branch): `node scripts/audit-evergreen-articles.mjs` → exit 0, `{"totalEvergreen":955,"staleCount":2,...}`; `--section=svizzera` → exit 0, `{"totalEvergreen":211,"staleCount":0,"stale":[]}`.

- **#3284** (data-loss bug, root cause fixed) — `scripts/scan-prev-slug-losses.mjs`'s watchdog flagged 6136 `previousSlugs` losses in 24h with none of it appearing in the slug-history-journal event log, proving an un-journaled mutation path. Root cause: `mergeAndDeduplicate()` in `scripts/lib/dedicated-crawler-common.mjs` (sole caller `scripts/lib/shared-jobs-crawler.mjs`) had 3 of its 4 duplicate-collision merge passes call `preferJob(a, b)` bare — `preferJob` returns one whole job object, never merges fields, so the loser record's independently-accumulated `previousSlugs`/`previousSlugsByLocale` history was silently discarded with zero journal entry. Fix: new `mergeDuplicateJobPreservingSlugHistory(a, b)` helper unions both sides' slug history (reusing the existing journal-backed `mergePreviousSlugsByLocale` + `captureLostSlugs` helpers) before `preferJob` picks a winner; applied at all 3 buggy call sites, the 4th (already correct) left untouched as reference. Sibling fixed in the same commit (AGENTS.md #6): `scripts/scatter-jobs-to-slices.mjs` had a second, independent journal-bypass — a hand-rolled push into `previousSlugsByLocale` plus a flat `previousSlugs` cap that kept the OLDEST 20 entries instead of the most recent, with no `recordSlugMutation` call — now routed through the canonical `addPreviousSlugForLocale` helper. 2 new regression tests added (isolated via `SLUG_REGISTRY_PATH_OVERRIDE`, never touches the tracked `data/slug-registry.json`). Re-verified on this bundle's HEAD after resolving a merge conflict in the shared test file (two independently-added test suites appended at the same location — combined both, no test dropped): `npx vitest run tests/dedicated-crawler-common.test.ts tests/slug-history-journal.test.ts` → 2 files, 33 tests, all passed.

Addresses item 3 (of 3) of #3206 — Firestore race in `maybeEscalateSoftBounce` (read-decide-write, called from all 5 ESP webhook cores) wrapped in a transaction. Sibling check found + fixed 2 more instances of the same unprotected construct on `newsletter_subscribers`: `refreshEngagementScore` and Resend webhook's status-promotion path. A distinct business-logic gap surfaced in the Resend path (promotion gate doesn't exclude terminal negative statuses) — not a concurrency bug, filed separately as #3305 rather than silently dropped. Items 1+2 of #3206 untouched (item 1 = owner-gated production `--apply`, item 2 already done in an earlier round).
Addresses item 1+2+3 (all items) of #2860 — item 1 (script-load `onerror` guard) already fixed by a pre-existing orphaned commit, verified and reused; item 2 (consent/auction/guard timeout reconciliation) fixed by pointing `consentManagement.gdpr.timeout` at the same 1000ms constant as `requestBids`, keeping consent wait inside the 1500ms wall-clock guard; item 3 (SPA bid-cache staleness) fixed with an ad-unit-code registry + reuse-guard, wired into the SPA's navigation hook as defense-in-depth. Full sibling audit across 5 other ad-tech call sites — all false positives, documented. #2860's title is a 3-item aggregate — orchestrator will manually shut it post-merge.

#3305 (filed above from #3206's sibling-check) is NOT in this PR's diff — an independent fix landed directly on `main` (`388d28690f2`, same `TERMINAL_NEGATIVE_STATUSES` gate + equivalent test coverage) before this PR's own fix was cherry-picked; the duplicate cherry-pick was aborted and #3305 was closed directly citing `388d28690f2`. No action needed here.
Addresses item 2 (of 2) of #3060 — item 1 already fixed by prior merged commit 8e3083840fd, item 2 now confirms sector pages structurally cannot be noindex (no `noindex` field in `SectorPageInputs`, hardcoded `robots: 'index,follow'`, thin-content gate drops pages pre-write rather than noindexing them); documented with comments + regression test rather than adding dead `noindexPaths` registration for a case that can't occur. Coverage gap also closed on item 1: none of the existing tests actually fed the order-independent noindex regex a `content`-before-`name` attribute string (every current emitter happens to write `name` first, so the reordering path merged in 8e3083840fd was never exercised) — added a dedicated regression test in `tests/job-market-snapshot.test.ts` covering both attribute orders (quoted + minifier-unquoted) plus negative cases (indexable robots meta, unrelated meta text containing "noindex"). #3060's title is a 2-item aggregate — both items are now done across two PRs, orchestrator will manually shut #3060 post-merge citing both commits.
Addresses #3173 — `scripts/audit-404-risk.mjs`'s served-set builder (`SHARD_REPOS`) missed the sibling of PR #3177 (merged 2026-06-30, Ticino job/company section carved into 4 dedicated GitHub Pages repos `frontaliere-ticino-{it,en,de,fr}`): live en/de/fr Ticino pages looked like 404s from 2026-07-01 onward (offender count ballooned from 19 reported to 3581 measured). Fixed by adding a `TICINO_SHARD_REPOS` clone loop mirroring the existing `SHARD_REPOS` loop, with its own `AUDIT_404_TICINO_SHARD_FLOOR[_<LOC>]` override and `meta.ticinoShards` surfaced in logs. Post-fix: sitemap coverage 0/398445 would-404 (was ~3560), newsletter hub-link offenders collapsed from ~600+ to 3 (all one company). That residual 3-offender gap was a genuinely separate bug — not folded into this fix, filed as new issue #3310, now also fixed below.
Closes #3310 — investigation refuted the issue's own "missing translation" hypothesis: `data/jobs/by-crawler/solothurner-spitaeler.json` has complete non-thin `descriptionByLocale` for all 157 jobs in it/en/de/fr, and the real per-locale company-hub loop in `jobsSeoPagesPlugin.ts` doesn't gate on job-level translation at all. Actual root cause: Solothurner Spitäler (soH) is canton-SO (no TI jobs), so it was never covered by the real per-canton hub — its only live page came from the legacy 404-recovery bridge (`build-plugins/companyHubBridgePlugin.ts`), whose `autoDiscoverCompanyHubs()` hardcoded `key = 'it::${slug}'` when seeding crawler-derived entries, leaving en/de/fr coverage undiscovered even though `data/evidence-index.json`/`data/cf-hot-404s.json` show real historical traffic to those exact URLs. Fix: loop over all 4 locales when seeding crawler-derived bridge entries, reusing the existing per-locale `COPY`/`buildHubPath` (no new content invented). Sibling-checked against `locationHubBridgePlugin.ts`/`jobOrphanBridgePlugin.ts`/`cfHot404BridgePlugin.ts` — none share the pattern, this was the only locale-asymmetric crawler-discovery loop.
Addresses item 1 (of 2), item 2 (of 2) of #3141 — both items done, orchestrator will manually shut #3141 post-merge citing this commit plus the earlier merged item-1 fix. Item 1 (slug-collision suffix stability) already fixed by prior merged commit `fd97f354ba3` (#3152), pinning the tie-break suffix to `event.id`. Item 2 (thin/near-duplicate content risk at scale — Lugano already at 63 events sharing near-identical templated copy, and 0/136 events have a source description since the tio-agenda crawler never captures one) fixed by folding each event's already-available `venue` field (44 distinct values among Lugano's 63) into the detail-page about-paragraph as a real per-page content differentiator, with a graceful fallback to the prior wording when venue is absent so no page regresses. 108 targeted tests green (events-detail-pages, events-pipeline, events-digest-article, events-nationwide-sources), including 3 new regression tests (venue appears in copy; two colliding events at different venues stay distinguishable; missing-venue events degrade cleanly with no dangling separator).

## Non implementato (ancora)

- #3104 item 1: reseed `data/text-html-ratio-baseline.json` — needs a live `seed-text-html-ratio-baseline.yml` dispatch with an `eventi`-inclusive `deploy_run_id` + committing the regenerated baseline; not a code change, orchestrator to dispatch post-merge.
- #3078 item 2: bundle-size-delta OOM measurement — blocked: needs a measured run (not verifiable pre-merge), tracked as its own follow-up per the original issue's suggested action.

## Test plan

- [x] Targeted vitest run across all touched areas (events pipeline/nationwide/digest/fb-scheduler, search-cluster shape guard, cross-canton drift bridge, related-search-clusters, mint-app-token, all 5 newsletter ESP webhook cores, ad-slot regression, undici local-LLM-fallback timeout enforcement, canton-off-target fill guard, canton-root-validity 301 guard, sitemap hreflang integrity (30331 cases), cluster-301 boilerplate-body, fuel-section classifier + deploy-workflow + article-section-classifier + inlanguage-whitelist): 30777/30777 passed.
- [x] `bash -n scripts/lib/deploy-it-pages-prep.sh` syntax check passed.
- [x] `app-auth-health-monitor.yml` validated via `python3 -c yaml.safe_load` + `actionlint`, both clean.
- [x] `seed-title-baselines.yml`/`seed-bfs-depth-baseline.yml`/`seed-orphan-pages-baseline.yml`/`seed-text-html-ratio-baseline.yml` validated via `yaml.safe_load` + `actionlint`, no new findings vs. baseline.
- [x] Handful of merge conflicts across the round, all resolved by keeping whichever side already had the correct, tested, more-current logic (never blind ours/theirs): `build-search-cluster-301-map.mjs`'s CLI main-guard between #2918-item2 and #3302 (kept the already-tested `invokedDirectly` guard); `newsletterResendWebhookCore.js`'s bounce-severity hoisting between #3206 and #2852 (kept #3206's hoisted vars, grafted in #2852's new reactivation call); #2918-item3's cron-wiring commit against its own item-2 predecessor in this same PR (pure docblock/export duplication — both sides already agreed on the actual logic, kept the more-accurate post-cron-wiring wording); `tests/dedicated-crawler-common.test.ts`'s import line + trailing test-suite append between #3089's slice-seeding tests and #3284's new dedup-slug-history tests (pure independent-addition collision at the same append point, no logic overlap — combined both import sets, kept both test suites in full, verified 33/33 pass afterward). #2918 item 1 (en/de/fr cluster-301 extension) is also already covered — shipped separately via merged PR #3300 (sibling issue #2923's identical scope). #2918 item 2's shape-guard duplicated an already-merged fix (`resolveLegacyUrlsArray`) — kept the existing, already-tested version, discarded the duplicate `extractIndexedUrlsArray`; kept the new, non-duplicate `assertNonEmptyLegacyMap` zero-map-after-match guard and the new prune script. All 3 items of #2918 are now complete.


valerielinc-ops:
⚠️ **collision-risk**: questa PR tocca file funnel-critical condivisi con la PR #3307: `scripts/lib/ai-models.mjs`. La seconda a raggiungere il merge DEVE prima rebasare oltre l'altra (`git merge origin/main` dopo che l'altra è mergiata) — l'auto-merge è bloccato finché `collision-risk` + dietro main. _Segnale deterministico da pr-collision-detector.yml (zero-Claude)._

valerielinc-ops (commented):
## Scope batch follow-up backlog round 3 — ~30 sub-fixes closing 7+ tracked issues across SEO redirect infra (CF Worker canton-root-validity guard), chunk-load-skew self-heal bootstrap byte-parity, svizzera-article canonical dedup, company-hub locale-symmetric 404 recovery, AI-model cache-key correctness, previousSlugs history preservation on job dedup, newsletter reactivation, and CI/workflow infra (App-token migration, >1GB Pages artifact download) (tier: high)

## Findings (Important: 0, Nit: 1)

- `scripts/lib/dedicated-crawler-common.mjs`: new `mergeDuplicateJobPreservingSlugHistory` (and the pre-existing `mergePreviousSlugsByLocale` it reuses) cap merged `previousSlugs` with `.slice(0, 20)` (keep-oldest) instead of the `addPreviousSlugForLocale` convention's `.slice(-cap)` (keep-most-recent) that this SAME PR adopts as the fix in `scripts/scatter-jobs-to-slices.mjs`. Pre-existing code (confirmed via `git log -S`, predates this PR) reused rather than newly introduced, and only manifests when two colliding duplicate records both exceed 20 combined unique previousSlugs for one locale — narrow/rare. Worth a follow-up to route it through the same `-cap` convention for consistency, not blocking.

Everything else sampled checks out on direct trace, not just the PR body's claims: `ogPagesPlugin.ts`'s canonical-override wiring for svizzera near-duplicates, `companyHubBridgePlugin.ts`'s 4-locale seed loop, the three-way byte-parity of the self-heal bootstrap (`services/resilientImport.ts` / `index.html` / `build-plugins/constants.ts`'s `SELF_HEAL_SCRIPT_CONTENT`, including safe regex-source interpolation), `ai-models.mjs`'s `getPreferredModel` + model-aware cache-key fix and its caller in `shared-jobs-crawler.mjs`, the new CF Worker `recoverPrunedCompanyHub`/`recoverExpiredJobToCantonRoot` guards (both fail closed to the existing soft-404 via `redirectTargetIsLive` on any miss/timeout, ordered correctly after the more-specific recoveries), `create-article.mjs`'s slug-collision guard now covering en/de/fr for both AI and journalist paths, and the new `tests/sitemap-slug-integrity.test.ts` hreflang-alternate checks deriving paths from the shared `articleSections.ts` registry instead of a duplicated literal. No multi-issue `Closes #a #b` violations, no `## Non implementato` items closed via deferral (both listed items have a real next step/blocked-cause per the completeness contract).

## Adversarial check
- `redirectTargetIsLive`'s CF Worker subrequest guard was only traced statically (fail-closed logic reads correctly); not exercised against a live shard's actual cache/latency behavior.
- The PR body's test-plan claim of 30777/30777 vitest passing was not independently re-run in this review pass.
- The self-heal bootstrap's regex-interpolated skew-detection was checked for textual/syntactic safety only, not exercised against a real chunk version-skew in a browser.

## LGTM
Standing in for the CI reviewer (5 consecutive quota-outage failures on `pr-review-loop.yml`, never posted). Reviewed the full code diff directly against the worktree at the PR head SHA, not just the PR body's claims. Zero Important findings; one narrow pre-existing nit noted for follow-up, not blocking.

PR #3307, opened 2026-07-02
fix: cap AI cascade wall-clock time, stop wasted timeout retries

valerielinc-ops:
## Implementato

Root-cause fix for run 28611052353 (Generate Blog Article, 109min / zero output). Diagnosed causes and fixes, all in `scripts/lib/ai-models.mjs` + `scripts/create-article.mjs`:

- **No retry on timeout within a single provider call** (`_callOpenAICompatible`, `_callGeminiRaw`): a hung model previously got retried once internally (2× full 90s waits = 180s per model). Observed 0/19 timeout retries succeeding in the diagnosed run, burning ~62min re-waiting on already-hung models. Now bails after the first timeout.
- **NVIDIA content-safety classifier models wrongly included in the chat cascade**: `nemotron-3-content-safety`, `nemotron-3.5-content-safety`, `nemotron-content-safety-reasoning-4b` matched the broad "nemotron" family allowlist regex but are moderation classifiers, not chat models — always fail with HTTP 400 "Conversation roles must alternate". Added to `NON_CHAT_MODEL_RE` (same pattern as the 2026-06-15 reward/parse fix).
- **Defense in depth**: `classifyNonRetryableError` now marks a model exhausted immediately (not just non-retryable) if the error is literally "Conversation roles must alternate". protects against a future model of this kind slipping past the name-based discovery filter.
- **Wall-clock budget gap**: `RUN_WALL_BUDGET_MS` (30min default) was only checked *between* outer headline-retry attempts, never inside a single `callLLM()` cascade walk. One walk of the ~180-model chain could (and did) consume nearly the entire budget before the outer check got a chance to fire. `ai-models.mjs`'s exported `callLLM()` now accepts an opt-in `deadlineMs` and bails out of the remaining chain once it's passed; `create-article.mjs` now threads its existing `RUN_START_MS + RUN_WALL_BUDGET_MS` into that param. Purely additive — no-op for the other 7 scripts that share this library and don't pass `deadlineMs`.
- Updated `tests/scripts/ai-models-dead-discovery-deny.test.ts`: removed `nvidia/nemotron-3.5-content-safety` from the "live sibling" allowlist fixture (it isn't a usable chat sibling — that assumption predates this fix) and added a dedicated test locking in the new content-safety exclusion. Full `tests/scripts/ai-models-*.test.ts` + `tests/local-llm-fallback.test.ts` suite green (50/50).
- **Review fix (round 1):** the deadline-skip message (`ai-models.mjs` inside the per-model loop) used "aborting" (present tense), which matched neither `transientRe` nor `persistentRe` in `classifyExhaustionCause`. When the deadline-skip is the *only* error in the cascade (deadline already past at `i===0` — exactly the case this PR targets, a prior `callLLM()` walk already burned the budget), that left the exhaustion cause unclassified → `isQuotaExhaustedError` returned `false` → `create-article.mjs`'s catch-all fell through to the hard-failure path (`finalizeRunReport('error', ...)` + `exit(1)`) instead of the intended graceful-defer branch, reproducing the exact false-positive Workflow-Failure this PR exists to prevent. Reworded to `"...deadline exceeded (timeout), aborted remaining chain..."` so it matches `transientRe`'s `timeout|aborted` tokens.

Sibling-pattern sweep (`node scripts/ci/check-sibling-patterns.mjs`) flagged 3 files sharing the `RUN_START_MS` token (`scripts/lib/translate-run-clock.mjs`, `scripts/local-mt-mopup.mjs`, `scripts/relocalize-pending-jobs.mjs`) — confirmed false positive: none call `ai-models.mjs`'s `callLLM()`, and their `RUN_START_MS`/deadline logic is an unrelated, already-complete per-script budget mechanism, not the "no deadline check inside the cascade walk" bug this PR fixes. No other site in the repo constructs a deadline-skip message (`grep -rn "deadline exceeded" scripts/ build-plugins/ services/` → single hit), so no further sweep needed for the review-fix wording.

## Non implementato (ancora)

Nessuno.

(The separate "explore local rewriting tools to reduce remote AI usage" ask from issue #3221 is intentionally out of this PR — it's an open-ended tool-evaluation task, not a bounded code diff, and will be handled as its own follow-up.)


valerielinc-ops:
⚠️ **collision-risk**: questa PR tocca file funnel-critical condivisi con la PR #3301: `scripts/lib/ai-models.mjs`. La seconda a raggiungere il merge DEVE prima rebasare oltre l'altra (`git merge origin/main` dopo che l'altra è mergiata) — l'auto-merge è bloccato finché `collision-risk` + dietro main. _Segnale deterministico da pr-collision-detector.yml (zero-Claude)._

PR #3348, opened 2026-07-03
feat(events): revamp hub/comune/detail UI + address+map card

valerielinc-ops:
## Implementato
- Event cards (hub + comune pages): category emoji + semantic-tone chip, image thumbnail when `event.imageUrl` is a mirrored site-relative path, gradient-tone placeholder otherwise.
- Hub and comune page hero headers restyled with gradient background + decorative emoji, exact existing copy text preserved.
- Comune grid cards on the hub page: hover-lift polish, arrow affordance.
- Event detail page: hero image (same mirrored-path guard) above the header; category chip upgraded to tone/emoji styling.
- New `renderLocationCard()` on the detail page: always shows venue/address text, embeds a plain OpenStreetMap iframe only when `event.geo` (real coordinates) is present — no fabricated pin, no Google Maps embed (same outbound-link policy as the existing `osmLink()` CTA, which the card now supersedes/replaces in the button row).
- Extracted the OSM bbox/iframe-src formula (previously duplicated only in `fuelDailyPagesPlugin.ts`) into a shared `osmEmbedSrc()` helper in `build-plugins/shared/seoContentTokens.ts`; `fuelDailyPagesPlugin.ts` now calls it too (verified byte-identical output — all 33 fuel-page tests pass unchanged).
- Updated `tests/events-detail-pages.test.ts` map assertions to match the new policy (iframe allowed with coordinates, still no Google Maps, still no fabricated map on baseline event with no coordinates).
- No structured-data (JSON-LD) changes — `Event`/`BreadcrumbList`/`FAQPage`/`ItemList`/`ImageObject` schema untouched, only visible markup/styling.
- No new translation keys — location-card heading reuses the existing translated `dc.whereLabel` copy in all 4 locales.

## Non implementato (ancora)
Nessuno.

Note: the "impeccable" design skill referenced in the original ask does not exist in this environment — the redesign instead follows the codebase's existing semantic-token conventions (`bg-X-subtle text-X border-X-border` tone pattern, confirmed via grep of existing usage) and Tailwind utilities already scanned by `tailwind.config.js`'s `./build-plugins/**/*.{js,ts}` content glob.

## Test plan
- [x] `npx vitest run tests/events-detail-pages.test.ts tests/events-pipeline.test.ts tests/fuel-daily-pages.test.ts` — 118 tests passing
- [x] `npx tsc --noEmit` — zero new type errors attributable to the 4 touched files (pre-existing unrelated errors elsewhere untouched)
- [x] `git diff origin/main --stat` scoped to only the 4 intended files after merging latest `origin/main`


valerielinc-ops:
Sibling-check triage (AGENTS.md #6): the OSM-embed construct I extracted (`osmEmbedSrc()` — static `embed.html?bbox=...&marker=...` iframe URL, no API key) has exactly one real sibling, `fuelDailyPagesPlugin.ts`, which this PR already refactored to use the shared helper.

`components/guide/BorderMunicipalitiesMap.tsx`'s "OpenStreetMap" hit is a **false positive**: it's a React-Leaflet `<TileLayer>` (interactive map, `tile.openstreetmap.org/{z}/{x}/{y}.png` tile URL + attribution string) — a completely different construct, no bbox/marker query-string building at all. Lexical match only (both say "OpenStreetMap"), semantically unrelated.

The remaining 40 candidates (`SiteEvent`, `imageUrl`, `mirrorEventImage`, `heroImage`, `cardTitle`, `renderEventCard`) are generic identifier-name collisions across unrelated domains (crawler scripts that already own/define `SiteEvent`/`mirrorEventImage` rather than duplicating them, and unrelated components like `TaxCalendar.tsx`/`BlogArticles.tsx`/`JobAlertForm.tsx` that happen to reuse common names for different purposes) — not instances of the pattern this PR touches.

PR #3374, opened 2026-07-03
fix(crawlers): wire itemprop microdata fallback for SF-CSB description extraction

### LINKED ISSUE #2823: Workflow Failure: Audit Parser Quality
**Workflow:** Audit Parser Quality

## Workflow fallito
**Run:** https://github.com/valerielinc-ops/frontaliere-si-o-no/actions/runs/28026685878
**Trigger:** workflow_run
**Ref:** main

valerielinc-ops:
## Implementato
- `readItemprop()` fallback wired into `parseCsbDetailPage` for Helsana's careers.helsana.ch (SuccessFactors CSB), which had dropped the `data-careersite-propertyid` attribute template — `readPropertyBlock(html, 'description')` was returning empty, and the ≥30-unique-word boilerplate guard was shipping synthesized filler instead of the real job description.
- Rewrote the tag-matching walk to track nested same-name tags to the true matching close tag, instead of stopping at the first inner `</span>` (the old regex truncated multi-paragraph descriptions to a few words).
- Extracted the shared logic into `locateTagByAttribute`/`extractBalancedTagBlock` in `scripts/lib/hospital-custom-html-helpers.mjs` so it isn't duplicated, and reused it to fix the identical truncation bug in `scripts/update-ist-jobs.mjs` (IST/Inselspital, same CSB template, same bug class per sibling-pattern rule).
- Added `tests/helsana-crawler.test.ts` (21 tests, passing).

Closes #2823

## Non implementato (ancora)
Nessuno.


valerielinc-ops:
Superseded by #3373 (round9-bundle), which already contains this exact fix (SF-CSB itemprop microdata fallback for Helsana/IST truncation) plus the Arsanté sibling fix already flagged and addressed there. Closing to avoid duplicate/conflicting merges — this PR's branch was built independently in a leftover worktree before I discovered #3373 was already in flight.

PR #3375, opened 2026-07-03
fix(seo): repair corrupted names + position drift in Articoli Frontaliere ItemList

### LINKED ISSUE #3358: seo-pages.ts ItemList: legacy position drift + duplicate positions (discovered via classifica-dogane-ticino registration)
While registering the new evergreen article `classifica-dogane-ticino` (border-wait ranking digest, weekly-refreshed), I found and fixed a real bug in `appendArticleListItem()` (scripts/lib/seo-pages-article-list.mjs): it derived the new ListItem `position` from the `numberOfItems` header text instead of the last item's own position, which can drift when items are added/removed ad-hoc over time. That fix landed together with the new article (see the PR this issue is linked from) and is covered by a regression test in `tests/scripts/seo-pages-article-list.test.ts`.

That fix is self-healing going forward, but it does not retroactively repair the **existing** drift already baked into `services/seo/seo-pages.ts`'s "Articoli Frontaliere" ItemList block:

- Real item count (~3059) vs. the max position label (~3087) — a ~28-item gap accumulated from past removals that never renumbered the block.
- At least two pre-existing duplicate-position pairs spotted while investigating: position 262 (near line ~4930) and position 951 (near line ~5600, whose `name` field also looked corrupted).

This is a dedicated renumbering/cleanup task on its own — touching thousands of lines of a large generated-looking data file — and is intentionally out of scope for the border-wait-ranking article PR (too invasive to bundle with an unrelated feature). Filing this so it's tracked rather than silently dropped.

Suggested approach: a one-off script that walks the ItemList block, renumbers `position` sequentially from 1, fixes the corrupted `name` at ~951, and rewrites `numberOfItems` to the true count — then a regression test asserting position == index+1 for every item and no duplicates.

valerielinc-ops:
## Implementato
- Repaired the "Articoli Frontaliere" ItemList JSON-LD block in `services/seo/seo-pages.ts`: 78 `ListItem` entries had a `name` field holding a raw, truncated JSON fragment (a prior generation step had serialized a whole object into `name` for some articles) instead of the real article title.
- Recovered 75 real titles by looking up the URL slug in `services/locales/blog-meta-it.ts`; removed 3 confirmed dead/orphaned entries with no recoverable title (`governo-tavolo-frontalieri-2026`, `kebab-case-3-5-words-max-40-chars`, `varese-soroptimist-studio-fibrosi-polmonare`).
- Fixed `position` collisions (two pairs of duplicate positions) and renumbered the whole block sequentially 1..N; corrected `numberOfItems` to match the true entry count (3056).
- Added `scripts/one-off/fix-itemlist-articoli-frontaliere.mjs`, an idempotent repair script (safe to re-run — no-op if the block is already clean), reusing the existing `locateItemListBlock` helper from `scripts/lib/seo-pages-article-list.mjs` rather than a second bounding implementation.
- Added `tests/regression/blog-articles-and-itemlist-dedup.test.ts` (5 tests, passing).
- Branch was several commits behind `origin/main` by the time this was ready to ship; re-ran the (idempotent) repair script against current data instead of hand-merging stale generated output — same corruption was still present on current `main`.

Closes #3358

## Non implementato (ancora)
Nessuno.


valerielinc-ops:
Superseded by #3373 (round9-bundle), which already contains this exact fix (Articoli Frontaliere ItemList repair, #3358) via the same idempotent one-off script. Closing to avoid duplicate/conflicting merges — this PR's branch was built independently in a leftover worktree before I discovered #3373 was already in flight.

PR #3622, opened 2026-07-06
fix(generate-article): resolve chained rebase-conflict data corruption + audit follow-up items (#3617, #3618)

### LINKED ISSUE #3617: follow-up(#3613): 1 item deferred — fix(generate-article): remove weak push-retry override (3→shared default 5)
## Origine
- PR: #3613 fix(generate-article): remove weak push-retry override (3→shared default 5) (merged 2026-07-06)
- URL: https://github.com/valerielinc-ops/frontaliere-si-o-no/pull/3613

### 1. Verify `resolve_append_conflicts` is idempotent across multiple rebase-retry cycles in the same push
- Source: reviewer adversarial check
- Original text:
  > `--in-place-resolver-cmd` (`resolve_append_conflicts`) is re-run on every one of the (now up to 4) rebase retries — not verified that repeated invocation across multiple rebase cycles in the same push is idempotent for all append-only file shapes it touches.
- Funnel impact: traffico (article generation pipeline — a non-idempotent resolver could corrupt or duplicate appended content across retry cycles, causing articles to be lost or malformed on push)
- Rationale: `generate-article.yml`'s push now retries up to 4 rebase cycles per push attempt (was 3). Each cycle re-runs `resolve_append_conflicts` via `--in-place-resolver-cmd`. If this helper appends without de-dup or guard, repeated invocations on the same conflict could accumulate duplicate entries in append-only files, corrupting the committed state. This sits in the article generation pipeline which directly drives content/traffic.
- Suggested action: Inspect `scripts/lib/git-push-with-retry.sh` and locate the `resolve_append_conflicts` resolver implementation. Verify it is idempotent when invoked multiple times per push (e.g. contains a de-dup check or is append-once-per-file-per-push). If not idempotent, add a guard (e.g. check whether the target content is already present before appending) or restructure the caller to run the resolver once per push attempt rather than once per rebase cycle.

valerielinc-ops:
## Implementato

**#3617 — resolve_append_conflicts fails on a second chained rebase-conflict cycle within one push retry loop**
- Root cause: since PR #3613 raised `git-push-with-retry.sh`'s default max-attempts 3→5, a push can now hit 2+ chained rebase conflicts on the same append-only JSON file within one retry loop. On the *second* cycle, each side's "keep both" hunk independently includes the container's closing token, leaving a duplicated root `}`/`]` that the old comma-only regex repair in `resolve_append_conflicts` can't fix — the whole push aborts and the generated article is lost (the exact failure #3613 tried to reduce, just shifted from cycle 4 to cycle 2).
- Fix: `scripts/lib/resolve-append-conflicts.sh` — added `stripPrematureRootCloses()`, a bracket-depth-tracking, string/escape-aware scanner that removes every root-depth closing token except the last one, before the existing comma-fixup regexes run. Nested closers untouched. Single shared implementation, so it covers all 3 real callers (`generate-article.yml`, `ai-visibility-check.yml`, `revenue-monitor.yml`).
- Verified empirically, not just reasoned: `tests/resolve-append-conflicts-idempotency.test.ts` drives real `git init`/`clone`/`rebase` against a temp bare repo, forcing two chained rebase conflicts on the same file within one push's retry loop, sourcing the actual shipped script. Confirms cycle 2 now resolves (previously failed exit 1), final push succeeds, JSON valid, no duplicate keys — plus a regression case for the ordinary single-cycle conflict.
- Follow-up bug found during this same investigation (pre-existing since the function's original PR #645, unrelated to #3613): the `ALL_BLOG_ARTICLE_IDS` merge in this same script spliced its merged replacement declaration into the string, then re-scanned the now-MUTATED string with the same global regex to strip the other duplicate — which also matched (and deleted) the declaration it had just inserted, silently wiping every id on every ordinary 2-way conflict on that line. Fixed by removing the other duplicates by their originally-captured offsets, back-to-front, never re-scanning after splicing in the replacement. New `tests/scripts/resolve-append-conflicts-rebase-idempotency.test.ts` locks this.

**Review follow-up (🔴 Important) — duplicated-root-closer repair extended beyond JSON**
- Reviewer found `stripPrematureRootCloses()` above only covers `.json`, while this script's own header already lists non-JSON append-only siblings sharing the same conflict pipeline that got no repair at all for the identical duplicated-root-closer corruption: `blog-articles-data.ts` (`RAW_ARTICLES`, closes with `] satisfies Article[];`), `swiss-articles-data.ts` (`RAW_SWISS_ARTICLES`, closes with plain `];`), `seo-pages.ts` (ItemList block), `blog-meta-*.ts` (`Record<string, string>` maps, closes with `};`), and `sitemap*.xml` (`<urlset>...</urlset>`) — the last of these had no check of any kind, not even the generic esbuild parse guard.
- Fix: new `scripts/lib/dedupe-growing-container-closer.mjs`, wired into `resolve-append-conflicts.sh` right after conflict-marker stripping (before the JSON/router/registry/esbuild steps). Each target is repaired via an anchor-bounded scan (`dedupeBounded`/`dedupeSelfBounded`) scoped to that container's own known open/close tokens — never a whole-file bracket scan — so it can't misfire on unrelated brackets elsewhere in a large file. Keeps only the last closer occurrence, drops the earlier (premature) ones.
- New `tests/dedupe-growing-container-closer.test.ts` (10 tests): one per target file class (both closing-token variants of the article arrays, the meta-map `};`, the sitemap `</urlset>`, the self-bounded ItemList), plus no-op cases for already-clean input and unrecognized files, and a CLI round-trip check.

Closes #3617

**#3618 item 1 — token-estimate divisor (chars/4 → chars/3.5) audited across the full model chain**
- Confirmed safe, no bug: swept every model in `DEFAULT_CHAIN` across chars 4000-35000 against each model's real `MODEL_MAX_REQUEST_TOKENS`/provider cap. Old and new divisor disagree only in narrow per-cap crossover windows; at those windows the tightened divisor is the one that's *correct* (verified against real tokenizer ground truth on the motivating incident, run 28732970228, ~29140 chars: old divisor under-estimated and would have let a doomed 413 through, new divisor correctly skips). No model is newly-and-incorrectly skipped.
- Change: documentation-only addition to `estimateRequestTokens`'s docstring in `scripts/lib/ai-models.mjs` recording the audit; no logic touched. New `tests/scripts/ai-models-token-estimate-divisor.test.ts` (5 tests) pins the formula and the exact crossover boundaries.

**#3618 item 2 — llm-json-repair asterisk/quote interaction**
- Already fixed in PR 3619 (which explicitly cites "issue #3618 item 2" in its own commit) — confirmed via the existing 40-test suite plus the issue's literal repro and 2 realistic variants, all parsing correctly. Added one permanent regression test to `tests/scripts/llm-json-repair.test.ts` using the exact `imagePrompt`/`imageAlt` field-pair shape from the issue (prior coverage used generic `a`/`b` keys).

Addresses items 1 and 2 of #3618 (both items verified complete; not using a closing keyword here because #3618's title matches this repo's aggregate-follow-up pattern ("2 item deferred") — `pr-body-contract.yml` blocks that pattern on multi-item aggregates by design, since it can't verify cross-PR completeness on its own. Issue 3618 will be marked done manually post-merge via a separate evidence-citing comment, per the same path used for issue 3598.)

**Sibling sweep (Non-Negotiable #6)** — `node scripts/ci/check-sibling-patterns.mjs` flagged 5 files sharing tokens with the diff; all 5 hand-confirmed false positives (independently re-verified, not just taken on the sub-agents' word):
- `scripts/seo-audit-structural.mjs` (`rootClose`/`rootCloseIdx`): unrelated HTML `</div>` string-index lookup, not JSON depth-tracking.
- `scripts/recover-conflict-marker-slices.mjs` (`inString`): a separate, manually-invoked recovery tool for a different incident (conflict-marker-left-in-committed-slice, not the auto rebase-retry-loop this PR fixes), with its own already-idempotent balanced-object + id-dedup design — never auto-invoked more than once per push.
- `.github/workflows/smoke-test-ai-models.yml`, `scripts/set-chutes-rc.mjs`, `scripts/smoke-test-ai-models.mjs` (`DEFAULT_CHAIN`): the `ai-models.mjs` diff is comment-only and never touches `DEFAULT_CHAIN`'s definition or logic — matched only because the new docstring prose mentions the identifier by name.

## Non implementato (ancora)
- Nessuno (entrambe le issue coperte: #3617 chiusa da `Closes` in questa PR; #3618 sarà chiusa manualmente post-merge dopo verifica live, essendo un'aggregata multi-item — vedi sopra)

### DIFF
--- a/scripts/lib/ai-models.mjs
+++ b/scripts/lib/ai-models.mjs
@@ -2471,6 +2471,30 @@ function _learnSchemaIncompatible(modelForTracking) {
  * and let a doomed HTTP call through. chars/3.5 + 500 on the same prompt
  * yields ~8826, within 1% of the real count and back over the cap.
  *
+ * Tightening chars/4 → chars/3.5 raises EVERY model's estimate uniformly, so
+ * issue #3618 item 1 asked whether this pushes some other DEFAULT_CHAIN
+ * model — one that was correctly passing before — into a false-positive skip
+ * at realistic prompt sizes. Audited by sweeping every model in DEFAULT_CHAIN
+ * across chars 4000..35000 (this codebase's real article-generation "Call
+ * 1/5" prompt, source text + instructions, is not a fixed 8-10k chars — the
+ * same run 28732970228 prompt that motivated this change was itself ~29140
+ * chars): old vs new divisor disagree ONLY inside a narrow per-cap window —
+ * (12251, 14000] chars for the 4000-token caps, (26251, 30000] chars for the
+ * 8000-token caps (the GitHub Models / Groq provider defaults that cover most
+ * of the chain). No model has a *different, lower* real limit than these —
+ * DEFAULT_REQUEST_TOKENS_BY_PROVIDER's own docstring notes the GitHub 8000
+ * cap was verified ACCOUNT-WIDE across five unrelated model families hitting
+ * byte-identical 413 text at the same estimate, i.e. token-counting for a
+ * given prompt doesn't vary by which model on that provider receives it — so
+ * a prompt landing in that window is genuinely at-or-over every model's real
+ * cap on that provider, not a false positive for some but not others. Net:
+ * no model is incorrectly newly-skipped; the tightened divisor only converts
+ * previously-silent false negatives (doomed 413/400 calls) into pre-flight
+ * skips, matching the fallback chain's own stated cost model (a wrong skip
+ * costs one chain step; a wrong pass costs a guaranteed failed HTTP call).
+ * See tests/scripts/ai-models-token-estimate-divisor.test.ts for the
+ * regression lock on the divisor value and the boundary math above.
+ *
  * Exported for tests / smoke probes.
  */
 export function estimateRequestTokens(messages, opts = {}) {
--- a/scripts/lib/dedupe-growing-container-closer.mjs
+++ b/scripts/lib/dedupe-growing-container-closer.mjs
@@ -1,0 +1,154 @@
+/**
+ * scripts/lib/dedupe-growing-container-closer.mjs
+ *
+ * Issue #3617 (JSON) found that a SECOND chained rebase-conflict cycle on
+ * the same append-only file within one push's retry loop (git-push-with-retry.sh
+ * --max-attempts, raised 3→5 by PR #3613) can produce a duplicated root
+ * closer: both sides' "keep both" hunks independently include the
+ * container's own closing token, so naive marker-stripping leaves that
+ * token twice with the other side's new entry orphaned in between.
+ *
+ * The SAME failure shape reaches several non-JSON append-only targets
+ * written by scripts/create-article.mjs / scripts/lib/seo-pages-article-list.mjs,
+ * because their insertion regexes anchor on the container's closing token
+ * itself (not on the preceding entry), so a second conflict on that same
+ * anchor duplicates it exactly like JSON did:
+ *   - data/blog-articles-data.ts   (`const RAW_ARTICLES = [ ... ]`)
+ *   - data/swiss-articles-data.ts  (`const RAW_SWISS_ARTICLES = [ ... ]`)
+ *   - services/seo/seo-pages.ts    ("Articoli Frontaliere" ItemList block)
+ *   - services/locales/blog-meta-{it,en,de,fr}.ts and blog-meta-ch-*.ts
+ *     (`Record<string, string>` meta-key objects)
+ *   - public/sitemap-blog.xml, sitemap-blog-ch.xml, sitemap-news.xml,
+ *     sitemap.xml (`<urlset>...</urlset>`)
+ *
+ * `services/routerBlogData.ts`/`services/routerSwissData.ts`'s BLOG_SLUGS /
+ * SWISS_SLUGS maps are NOT in this list: their insertion anchors on the
+ * PRECEDING entry line, never touching the map's closing `};`, so two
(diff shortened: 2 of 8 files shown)


valerielinc-ops:
Adding an additional, independently-found fix to this PR since it already targets #3617/#3618 (avoiding a duplicate competing PR for the same issues).

**Additional fix (commit `f923b6dfc13`):** `resolve_append_conflicts()`'s `ALL_BLOG_ARTICLE_IDS` dedup/merge block computed a correct merged declaration, then ran `src.replace(listRx, '')` with the same global regex to strip the leftover duplicate declarations — but that call re-scans the now-mutated string and matches the just-inserted merged replacement too, silently wiping the entire blog-article-id registry on any ordinary 2-way conflict (present since the function's original PR #645, confirmed via `git log -S`, untouched by 3 later hardening PRs including this one's `stripPrematureRootCloses()`).

Fixed by removing the non-first duplicate declarations via their originally-captured offsets (back-to-front, so earlier removals never shift not-yet-processed offsets), then splicing the merged replacement into the first offset — never re-scanning mutated text.

This is a distinct bug from `stripPrematureRootCloses()` above it in the same function (different section, different failure mode — root-closer duplication vs. id-registry wipe) — verified no line/logic overlap, and both fixes' test suites pass together (`tests/resolve-append-conflicts-idempotency.test.ts` + `tests/scripts/resolve-append-conflicts-rebase-idempotency.test.ts` → 4/4 passed).

No PR body change needed — this stays scoped under the existing `Closes #3617`.

PR #3684, opened 2026-07-06
Fix local-fallback article generation stall (ground organic mode, route local-only to evergreen)

valerielinc-ops:
## Implementato

- **Local-only cascade → evergreen routing** (`scripts/create-article.mjs`, `main()`): when the cloud model cascade is fully exhausted and only `local/fallback` (qwen2.5:14b) remains available, generation now routes directly to evergreen mode instead of attempting organic/news generation. Detection uses `getPreferredModel({chain: cloudOnlyChain})` — a zero-API-call peek at cascade state — and reuses the pre-existing `FORCE_EVERGREEN`-gated code path (`forceEvergreen` flag), so no new branching logic was added.
- **Grounding facts injected into organic/news prompts** (`callGemini` in `scripts/create-article.mjs`): `EVERGREEN_FACTS_BRIEF` (imposta alla fonte, franchigia, aliquote CH/IT, istituzioni valide, etc.) is now injected into the organic/news generation prompt too, not just evergreen mode. This closes the asymmetry where the fact-check gate (`llmFactCheck`) grades against `VERIFIED_DOMAIN_FACTS` but the generator had no access to those facts — the weak local model was fabricating domain facts that didn't match ground truth, which `llmFactCheck`'s "coerenza" dimension correctly flags as critical (fabricated facts) rather than minor (correct enrichment). Benefits remote models on regen attempts too, not just local.
- **Regeneration token budget tightened** (`callGemini` + regen-feedback loop in `main()`): `MAX_SOURCE_CHARS` now scales down on retry attempts (6000 → 4500 chars from attempt 2 onward), and per-issue fact-check feedback truncation tightened from 90/110 to 70/90 chars (issue count cap unchanged at 8). Keeps regen prompts under the ~8000-token cap that was excluding otherwise-eligible remote models on retries.
- `isLocalLlmEnabled()` in `scripts/lib/ai-models.mjs` exported (was module-private) so `create-article.mjs` reuses the same truthy-check instead of duplicating the regex.

Root cause addressed: zero articles produced across 34+ commits to `main` since the qwen2.5:14b local-fallback bump (PR #3642) — organic-mode generation with the local model had no grounding, so it hallucinated domain facts that the fact-check gate correctly blocked as critical. Verified via GitHub Actions run history that this was a genuine production stall, not a false alarm.

## Non implementato (ancora)

Nessuno. Deferred items identified during exploration (local RAG/retrieval via onnxruntime-node, template/slot-based evergreen generation, closing the fact-check fail-open hole) were explicitly out of scope for this fix and not requested — not part of this task's scope, so not tracked here as open work.

## Test plan

- [x] `npx vitest run tests/create-article-gates.test.ts tests/scripts/ai-models-get-preferred-model.test.ts tests/scripts/create-article-integration.test.ts tests/fact-check-consensus.test.ts` — 78/78 passing, re-verified after merging latest `origin/main`.
- [x] `node --check` on both edited files — syntax valid.
- [x] Sibling-pattern sweep (`scripts/ci/check-sibling-patterns.mjs`) — 18 flagged files manually reviewed, all false positives (lexically similar names or shared-import co-occurrence, not the same construct).
- [x] GitNexus impact/context on `callGemini`/`generateAndValidateArticle` — single internal caller, LOW risk, narrow blast radius.
- [ ] Live verification: next `generate-article.yml` run on `main` actually produces a published article (to be checked post-merge).

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/scripts/create-article.mjs
+++ b/scripts/create-article.mjs
@@ -65,7 +65,7 @@ import { execSync } from 'node:child_process';
 import { createInterface } from 'node:readline';
 import path from 'node:path';
 import { pathToFileURL } from 'node:url';
-import { callLLM as _aiCallLLM, AI_MODELS, getStats as getAiStats, initScoreStore, flushScores, recordModelContentFailure, recordModelContentSuccess, isQuotaExhaustedError, printRunSummary } from './lib/ai-models.mjs';
+import { callLLM as _aiCallLLM, AI_MODELS, DEFAULT_CHAIN, getPreferredModel, isLocalLlmEnabled, getStats as getAiStats, initScoreStore, flushScores, recordModelContentFailure, recordModelContentSuccess, isQuotaExhaustedError, printRunSummary } from './lib/ai-models.mjs';
 // Quota-free MT cascade (DeepL-free / Google / MyMemory / LibreTranslate /
 // local Opus-MT) — the SAME translator the job crawlers + FAQ batch use
 // (scripts/lib/dedicated-crawler-common.mjs, batch-add-faq-to-articles.mjs).
@@ -4122,7 +4122,14 @@ async function callGemini(pageContent, url, sourceContext = null) {
   //   2. Provide generous source content (6000 chars) so the model has facts to work with
   //   3. Send compact IT-only JSON template (EN/DE/FR generated in separate calls)
   //   4. Compress editorial rules (no repetition per locale)
-  const MAX_SOURCE_CHARS = 6000;
+  const generationAttempt = Number(sourceContext?._generationAttempt || 1);
+  // Regen attempts (2+) also carry factCheckRefinementInstruction (flagged
+  // claims to fix) and, since fix B above, domainFactsBlock — both compete
+  // with source content for the same ~8000-token input cap several free
+  // models enforce. Shrinking the re-sent source on retries only (never on
+  // the first, richness-matters attempt) buys headroom without touching
+  // first-attempt grounding.
+  const MAX_SOURCE_CHARS = generationAttempt > 1 ? 4500 : 6000;
   const MAX_IDS_TO_SEND = 50;
 
   const truncatedContent = pageContent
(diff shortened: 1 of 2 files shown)


valerielinc-ops (commented):
**Local review (remote reviewer stuck in runner backlog — 30+ min queued, systemic saturation across the repo). Applying REVIEW.md contract manually per owner instruction.**

## Scope
Route auto-scan article generation to evergreen mode when only local/fallback survives the cloud cascade, inject grounding facts into organic/news prompts too, and tighten regen-attempt token budgets (tier: high).

## Findings (Important: 0, Nit: 2)

- `scripts/create-article.mjs` — 🟡 Nit: Phase-1 comment ("skipped only on explicit FORCE_EVERGREEN=1") is now stale — `forceEvergreen` can also be true via `cloudCascadeExhausted`. Update comment. — non funnel-critical.
- `scripts/create-article.mjs` (callGemini) — 🟡 Nit: comment claims both `evergreen://` and `stats-bfs://` branches bake `EVERGREEN_FACTS_BRIEF` into `pageContent` upstream, but the stats-bfs path actually embeds concrete BFS trend numbers via `buildStatsBfsPromptContent`, not the brief constant. Harmless (topic doesn't need the tax brief) but comment overstates. — non funnel-critical.

Completeness contract: both headers present, `## Non implementato (ancora)` = "Nessuno" with reasoned deferred-scope list — compliant.

Traced and verified: `isSyntheticSource` covers all 3 url schemes reaching `callGemini` (single caller, gated input); `generationAttempt` relocation is required (const-before-use), no duplicate left; `initScoreStore()` explicit call is necessary and idempotent, no ordering hazard vs the old lazy auto-init; facts brief stays well under the tightened 4500-char regen cap; fact-check feedback truncation (70/90 vs 90/110) is free-text LLM guidance only, cosmetic. `node --check` clean on both files.

## Adversarial check
1. ❓ `domainFactsBlock` (~1650 chars) added unconditionally on organic attempt 1 too, uncounted against `MAX_SOURCE_CHARS` — could shrink viable-model set for smaller-context free models on first attempt. Degrades gracefully (callLLM skips over-limit models), not funnel-breaking.
2. ❓ PR body's root-cause claim (zero articles across 34+ commits) taken on trust from the author's investigation, not independently re-verified here.
3. ❓ Whether the weak local model reliably honors "usali solo se pertinenti al tema" and doesn't bleed unrelated domain facts into off-topic articles — mitigated by existing pre-LLM topical gate, not traced end-to-end.

None escalate — all have graceful degradation paths.

## LGTM

PR #3701, opened 2026-07-06
feat(crawlers): consolidate 581 individual workflows into 23 grouped workflows

valerielinc-ops:
## Implementato

- Replaced 581 individual `.github/workflows/update-jobs-*.yml` crawler workflows with **23 consolidated `.github/workflows/crawler-group-{01..23}.yml`** workflows. Each dispatched individual workflow used to hold a full GitHub Free-tier concurrent-job slot (20 max) for its entire run duration (mean ~27min, up to ~160min for Coop; ~1160 dispatches/day), starving other CI (PR tests, `pr-review-loop`) of runner slots. Each group now runs its member crawlers as GitHub Actions "parallel steps" (`background: true` + a final `wait-all: true` step) — every crawler in a group executes concurrently inside ONE job, so a group holds exactly one concurrent-job slot regardless of member count.
- New `scripts/generate-crawler-group-workflows.mjs` (the generator, checked into the repo, re-runnable). Reads `data/crawler-manifest.json` (per-crawler step manifest: install/prep steps, run step, housekeeping/commit/error-report steps, extracted verbatim from the original 581 workflow files) and `data/crawler-workflow-duration-baseline.json` (historical average durations). Bin-packs crawlers into 23 groups with a custom algorithm (`packGroups`, exported and unit-tested) — group wall-clock = MAX member duration, not sum, since background steps run concurrently: it isolates genuine duration outliers (Coop, ~160min, ~6.5x the corpus median) into a singleton group, then anchors the next-longest items one per group and spreads the remaining tail evenly by member count. A naive "always add to the lowest-current-wallClockMs group" greedy was tried first and found to degenerate (22 near-empty groups + 1 group of 559) — documented in the function's own comments and guarded by tests.
- For each crawler, its entire original step sequence (run → housekeeping → commit+push → failure-report) is collapsed into ONE composite shell script per background step — GitHub's `background: true` applies to a single `run:` block, not a multi-step sub-job. Success/failure gating (`if [ "$crawler_exit" -eq 0 ]` around housekeeping/commit, `if [ "$crawler_exit" -ne 0 ]` around the failure report) reproduces the exact semantics of the original per-crawler workflows, where a failing run step (no `continue-on-error`) halted subsequent ungated steps.
- **slug-history-journal collision fix**: `scripts/lib/slug-history-journal.mjs`'s `defaultSummaryPath()` writes per-run telemetry to `SLUG_HISTORY_SUMMARY_FILE || /tmp/slug-history-summary-${pid}.txt`; `scripts/lib/git-commit-data.sh` falls back to `ls -t /tmp/slug-history-summary-*.txt | head -1` (globally-newest file, no crawler-name binding) when unset, then deletes it. Once many crawlers run as concurrent background steps sharing one job's `/tmp`, a crawler's commit step could steal + delete a sibling's telemetry file (misattributed commit message + silent loss). Fix: every crawler's composite step sets `SLUG_HISTORY_SUMMARY_FILE=/tmp/slug-history-summary-<slug>.txt` (unique per crawler name) **at the generated-YAML callsite only** — `slug-history-journal.mjs` and `git-commit-data.sh` themselves are untouched, since their PID-based default + glob fallback remain correct for any future single-crawler-per-job usage (e.g. manual local dispatch).
- **git-index race fix**: `git-commit-data.sh` runs `git add`/`git commit` directly against the shared working-copy `.git/index` with no locking — safe when each crawler is its own separate runner/clone, unsafe when several background steps in one job share the same working directory. Fix: each crawler's commit-and-push invocation is wrapped in `flock /tmp/crawler-group-git.lock -c '...'`, serializing only the few-second commit moment across a group's concurrent siblings; the crawl/fetch work itself still runs fully concurrently.
- Per-crawler commit-and-push and error-reporting mechanisms were left **entirely untouched by design** — no shared/generic commit step or shared/generic error-reporting step was introduced. Every crawler still commits and reports errors via its own extracted steps, verbatim, just co-located in one job instead of its own workflow run.
- Updated `.github/workflows/orchestrate-crawlers.yml` to discover/dispatch the 23 `crawler-group-*.yml` files instead of 581 `update-jobs-*.yml` files. Kept the dispatch-verification/polling logic (fire, poll for the new run, retry up to 3x) and the zombie-queued-run reaper unchanged; removed the now-unnecessary 2-way job sharding (existed to fit 581 sequential dispatches inside a 180min timeout — 23 targets finish in minutes) and simplified the volume-tiered dispatch delay (Large/Medium/Small) to a flat, short delay (default 20s, configurable) since 23 targets don't need 581's anti-thundering-herd spacing.
- Updated `scripts/scaffold-crawler.mjs`: no longer generates a standalone per-crawler workflow file. It now parses the same per-crawler workflow-step template it already generated (unchanged in content) into the manifest-entry shape and upserts it into `data/crawler-manifest.json`; the printed next-steps guide tells the operator to run `node scripts/generate-crawler-group-workflows.mjs` afterwards to fold the new crawler into a group. Verified end-to-end with a throwaway synthetic test crawler (manifest upsert + group regeneration), then reverted the test artifacts before committing.
- Fixed `build-plugins/crawlerRegistryPlugin.ts` (feeds the admin panel's crawler-dispatch registry, `dist/data/jobs-crawler-workflows.json`), which scanned `update-jobs-*.yml` and would have silently returned zero crawlers post-deletion. Now scans `crawler-group-*.yml` (via each background step's stable `id: crawler-<slug>` marker) and emits one registry entry per member crawler, each still carrying its own `slug`/`summaryKey`/title, with `id` now pointing at its containing GROUP workflow file (the only thing GitHub can actually dispatch/poll) plus new `groupId`/`groupMemberCount` fields.
- Fixed `components/pages/AdminPanel.tsx` to work against the new registry shape: a React list-key collision (multiple crawlers sharing the same group `id` would previously collide on the same key, silently dropping rows) is fixed by keying on the crawler's own `slug` instead; the admin panel's failed-step display is narrowed to the crawler's own `Run <slug>` step (via substring match — see Non implementato below for a known limitation) instead of showing every sibling's failures in the group's shared job-run state.
- Added `tests/generate-crawler-group-workflows.test.ts` (10 tests): 8 on `packGroups` directly (exactly 23 groups produced, every crawler in exactly one group, the real-corpus regression scenario that caught the original bin-packing bug, outlier isolation, max-not-sum wall-clock, safety-ceiling margin, fewer-crawlers-than-groups edge case, determinism), plus 2 end-to-end tests on `generate()` itself using a synthetic manifest+baseline in a temp dir, added during this review round to guard the `--ignore-scripts` propagation fix (see below). These tests assert on the generated YAML/shell-script text; they do not execute the rendered bash under a real shell.
- Fixed `tests/git-commit-data-auth.test.ts`, which asserted against 3 specific now-deleted `update-jobs-*.yml` files (their `github.token`-explicit auth hardening for hoch-health, hopital-de-lavaux, spital-lachen). Rewrote to locate each crawler's background step dynamically by its `name: Run <slug>` marker inside whichever `crawler-group-*.yml` currently contains it (not hardcoded to a specific group number, since bin-packing can reassign groups on regeneration).
- Updated `docs/CRAWLERS.md`, `docs/CI-CD-PIPELINE.md` (Stage 2/3 + the Phase-2 concurrency-impact section), `docs/CATHEDRAL-STATUS.md`, `docs/CATHEDRAL-IMPLEMENTATION-PLAN.md`, `docs/HOSPITAL-CRAWLERS-INVENTORY-2026-05-19.md` to describe the new architecture and fix now-broken instructions that referenced deleted per-crawler workflow files (e.g. "copy `update-jobs-bethesda-spital.yml`").
- Updated remaining prose comments referencing the `update-jobs-*` naming convention in `scripts/lib/crawler-template.mjs` and `scripts/lib/github-issue-creator.mjs` (descriptive only, no functional pattern-matching on the string) — grepped the rest of the codebase for the same pattern and confirmed no other functional callsite depends on it.
- **Review-round fix**: the shared "Install dependencies" step was hardcoded to plain `npm ci` for the whole group job, but 10 crawlers' original workflows used `npm ci --ignore-scripts`, and that flag was silently dropped once grouped — worse, since the step has no `continue-on-error`, a flaky dependency postinstall script failing there would have blocked every background crawler step in that group from starting, not just the one crawler needing the flag. Fixed: the shared install step now uses `--ignore-scripts` whenever any group member requires it (safe superset). Regenerated the 7 affected group workflows (04, 07, 11, 13, 15, 18, 19) and added the 2 end-to-end tests referenced above.
- Exact file-count change: **581 workflow files deleted, 23 added** (net -558 `.github/workflows/*.yml` files), plus 2 new committed data files (`data/crawler-manifest.json`, `data/crawler-workflow-duration-baseline.json`) and the new generator script + test file.
- `data/crawler-manifest.json` and `data/crawler-workflow-duration-baseline.json` were produced by one-off extraction scripts run locally during implementation (bulk `grep`/`yaml`-parse of all 581 original workflow files for the manifest; a paginated `gh api repos/{owner}/{repo}/actions/runs` query, filtered to `update-jobs-*` workflow paths and successful `workflow_dispatch` runs, for the duration baseline) — those one-off scripts were **not committed**; only their JSON output was, since the generator only needs to read the committed data files, not re-run the extraction. `yaml` is already an existing transitive dependency in this repo's `package-lock.json` (via other tooling) and was used as-is for both the one-off extraction and the generator's own output validation — it was not newly added as a direct devDependency.

## Non implementato (ancora)

Nessuno.

Honest caveats (not gaps, documented limitations):
- Each group's estimated wall-clock is a bin-packing estimate from **historical average durations** (one paginated GH API pull covering each crawler's most recent successful runs). Actual per-run duration varies (site slowness, job-count spikes, network conditions, retry backoffs) — the balancing minimizes the expected worst case, it does not bound the actual one. The 340-minute job timeout (vs GitHub's 360-minute hard cap) is the safety margin for that variance. If a specific group's real-world wall-clock consistently runs hot against this margin post-merge, re-running the generator with a refreshed baseline rebalances it — routine maintenance, not a defect.
- `AdminPanel.tsx`'s failed-step narrowing (`ownStepName` matched via `allFailedSteps.filter(s => s.includes(ownStepName))`) is a substring match, not an exact/anchored match — if one crawler's slug is a prefix of another's in the same group (e.g. `Run hoch-health` is a substring of a hypothetical `Run hoch-health-x`), the sibling's failed steps could still show up for the wrong crawler in the admin panel. Confirmed real by the reviewer, tagged non-funnel-critical (admin-UI display only, not data correctness) — deferred, not fixed in this PR; see the inline reply on the review thread.

### Group manifest summary

| Group | Members | Est. wall-clock |
|---|---|---|
| crawler-group-01 | 1 (Coop, isolated outlier) | ~160min |
| crawler-group-02 | 26 | ~86min |
| crawler-group-03 | 26 | ~77min |
| crawler-group-04 | 26 | ~72min |
| crawler-group-05 | 26 | ~71min |
| crawler-group-06 | 26 | ~66min |
| crawler-group-07 | 26 | ~66min |
| crawler-group-08 | 26 | ~64min |
| crawler-group-09 | 26 | ~63min |
| crawler-group-10 | 26 | ~63min |
| crawler-group-11 | 26 | ~61min |
| crawler-group-12 | 26 | ~60min |
| crawler-group-13 | 26 | ~60min |
| crawler-group-14 | 26 | ~59min |
| crawler-group-15 | 26 | ~59min |
| crawler-group-16 | 27 | ~58min |
| crawler-group-17 | 27 | ~58min |
| crawler-group-18 | 27 | ~58min |
| crawler-group-19 | 27 | ~57min |
| crawler-group-20 | 27 | ~57min |
| crawler-group-21 | 27 | ~56min |
| crawler-group-22 | 27 | ~56min |
| crawler-group-23 | 27 | ~56min |

Total: 581 crawlers across 23 groups.

## Test plan

- [x] `npx vitest run tests/generate-crawler-group-workflows.test.ts tests/git-commit-data-auth.test.ts` — 13/13 green
- [x] Full local suite (`node scripts/assemble-jobs-dataset.mjs --stats && npm test`) — 1208 test files passed, 106312 tests passed, 35 skipped (pre-existing, unrelated), 0 failed
- [x] All 23 generated `crawler-group-*.yml` files parse as valid YAML; every crawler's inlined background-step shell body passes `bash -n` syntax check (581/581)
- [x] Generator is deterministic (re-run produces byte-identical output — verified via `git status`/`git diff` showing no drift after regeneration)
- [x] `scaffold-crawler.mjs` end-to-end tested with a synthetic test crawler (manifest upsert + group regeneration verified), test artifacts reverted before commit
- [x] `crawlerRegistryPlugin.ts` extraction spot-checked against the real generated files
- [ ] Live orchestrator dispatch of the 23 groups (first real cron run, or a manual `gh workflow run orchestrate-crawlers.yml`) — not run as part of this PR per task instructions (no live crawler run without an explicit ask); will observe on `main` post-merge at the next scheduled 09:00/21:00 UTC run.


valerielinc-ops:
Re: reviewer nit on `components/pages/AdminPanel.tsx:2115-2118` (substring match on `ownStepName` instead of exact/anchored match, could show a sibling's failed steps for a crawler whose slug is a prefix of another's in the same group) — confirmed real, deferred — non funnel-critical. This affects only the admin panel's failed-step display text (a debugging aid for operators), not any data correctness, commit behavior, or user-facing output. Left as-is per repo policy for dropping confirmed-non-blocking nits without opening an issue; noted in the PR body's `## Non implementato (ancora)` caveats section.

PR #3703, opened 2026-07-06
Consolidate 581 crawler workflows into 23 grouped workflows

valerielinc-ops:
## Implementato

- Replaced 581 individual `.github/workflows/update-jobs-*.yml` crawler workflows with **23 consolidated `.github/workflows/crawler-group-{01..23}.yml`** workflows. Each dispatched individual workflow used to hold a full GitHub Free-tier concurrent-job slot (20 max) for its entire run duration (mean ~27min, up to ~3h for Coop; ~1160 dispatches/day), starving other CI (PR tests, `pr-review-loop`) of runner slots. Each group now runs its member crawlers as GitHub Actions "parallel steps" (`background: true` + `wait-all: true`) — many crawlers execute concurrently inside ONE job, so a group holds exactly one concurrent-job slot regardless of member count.
- `scripts/extract-crawler-manifest.mjs` — bulk-parses (via the `yaml` package, added as an explicit devDependency) every crawler's own bespoke steps (install, optional Playwright, Firebase prep, RC-secrets load, run-crawler, scoped housekeeping, commit+push, report-failure) into the committed `data/crawler-manifest.json`. Real inspection of all 581 files (not assumed) found meaningful per-crawler variance beyond commit-message text — some pass an extra `data/jobs-crawler-adapters/` path to `git-commit-data.sh` (526/581), 10 use `npm ci --ignore-scripts`, 30 install Playwright + use `xvfb-run -a`, 2/581 lack a housekeeping step, step names differ ("Housekeeping" vs "Scoped housekeeping"). The manifest preserves every crawler's own steps VERBATIM instead of reconstructing them from a generic template.
- `scripts/fetch-crawler-workflow-durations.mjs` — pulls real recent successful-run durations per crawler via the GitHub REST API (bulk per-workflow calls, not a single paginated scan of the repo's 144k+ runs), writes the committed `data/crawler-workflow-duration-baseline.json`. All 581 workflows had usable history (0 fallback-to-median needed in this run); falls back to the corpus median for any crawler with no run history.
- `scripts/generate-crawler-group-workflows.mjs` — the generator. Bin-packs crawlers into 23 groups via the repo's existing LPT (longest-processing-time) bin-packer (`scripts/ci/lpt-shard.mjs`, already used to balance vitest shards — reused rather than reimplemented), using MAX (not sum) duration per group since background steps run concurrently. Isolates extreme outliers (Coop, ~177min measured) into a dedicated group first. For each crawler, collapses its extracted bespoke steps into ONE composite shell script (GitHub's `background: true` applies to a single `run:` block, not a multi-step sub-job) with **per-sub-step subshell isolation** — a regression I caught and fixed during implementation: naively concatenating all steps into one shell process would let an early `exit 0` inside e.g. the real "no Firebase secret → skip" branch (present in 580/581 crawlers) silently terminate the WHOLE composite, skipping every later sub-step (Run crawler, Commit, Report-failure) for that crawler. Verified against the real corpus and with executable bash tests (`tests/generate-crawler-group-workflows.test.ts` actually runs the rendered scripts, not just string-matches them).
- **slug-history-journal collision fix**: `scripts/lib/slug-history-journal.mjs`'s `defaultSummaryPath()` writes per-run telemetry to `SLUG_HISTORY_SUMMARY_FILE || /tmp/slug-history-summary-${pid}.txt`; `scripts/lib/git-commit-data.sh` falls back to `ls -t /tmp/slug-history-summary-*.txt | head -1` (globally-newest file, no crawler-name binding) when unset, then deletes it. Once many crawlers run as concurrent background steps sharing one job's `/tmp`, a crawler's commit step could steal + delete a sibling's telemetry file (misattributed commit message + silent loss). Fix: every crawler's composite step sets `SLUG_HISTORY_SUMMARY_FILE=/tmp/slug-history-summary-{crawlerSlug}.txt` (unique per crawler name) **at the generated-YAML callsite only** — `slug-history-journal.mjs` and `git-commit-data.sh` themselves are untouched, since their PID-based default + glob fallback remain correct for any future single-crawler-per-job usage (e.g. manual local dispatch).
- Per-crawler commit-and-push and error-reporting mechanisms were left **entirely untouched by design** — no shared/generic commit step or shared/generic error-reporting step was introduced anywhere. Every crawler still commits and reports errors via its own extracted steps, verbatim, just co-located in one job instead of its own workflow run.
- Updated `.github/workflows/orchestrate-crawlers.yml` to discover/dispatch the 23 `crawler-group-*.yml` files instead of 581 `update-jobs-*.yml` files. Kept the dispatch-verification/polling logic (fire, poll for the new run, retry up to 3x) and the zombie-queued-run reaper unchanged; removed the now-unnecessary 2-way job sharding (existed because a single sequential pass over 425+ workflows couldn't finish inside a 180min timeout — 23 targets finish in minutes) and simplified the volume-tiered dispatch delay (Large/Medium/Small) to a flat, short delay (default 20s) since 23 targets don't need 581's anti-thundering-herd spacing. Restored the orchestrator's forced `skip_ai_translation=1` override on every dispatch (translation runs centrally via `translate-pending.yml`) which I'd initially dropped and caught in review of my own diff.
- Updated `scripts/scaffold-crawler.mjs`: no longer generates a standalone per-crawler workflow file. It now appends a manifest entry to `data/crawler-manifest.json` (via the new shared `scripts/lib/crawler-manifest-entry.mjs` step-shape builder — kept in one module so the manifest format can't drift between the bulk extractor and the scaffolder) and automatically re-runs the group generator, folding the new crawler into its least-loaded group. Verified end-to-end with two synthetic test crawlers (fresh append + `--force` overwrite), then reverted the test artifacts.
- Fixed `build-plugins/crawlerRegistryPlugin.ts` (admin UI's crawler-dispatch registry, `dist/data/jobs-crawler-workflows.json`) which scanned `update-jobs-*.yml` and would have silently returned zero crawlers post-deletion. Now scans `crawler-group-*.yml` and extracts one registry entry per member crawler (verified: 581/581 extracted correctly across all 23 files), each still carrying its own `summaryKey`/title for the "last crawl summary" lookup, honestly labeled that dispatching it runs its whole group (`groupSize` field) since individual crawler dispatch no longer exists.
- Added `tests/generate-crawler-group-workflows.test.ts` (16 tests): disjoint-cover invariant (every crawler in exactly one group), outlier isolation, LPT-vs-naive balance, determinism, and — critically — actually **executes** the rendered composite bash scripts (not just string assertions) to prove the subshell-isolation fix, `$GITHUB_ENV` propagation across the subshell boundary, and continue-on-error/if:failure() semantics all hold at runtime.
- Fixed `tests/git-commit-data-auth.test.ts`, which asserted against 3 specific now-deleted `update-jobs-*.yml` files (their `github.token`-explicit auth hardening); rewrote to assert against `data/crawler-manifest.json` instead (durable source, decoupled from which group each crawler happens to land in).
- Updated `docs/CRAWLERS.md`, `docs/CI-CD-PIPELINE.md` (Stage 2/3 + Phase-2 concurrency-impact section), `docs/CATHEDRAL-STATUS.md`, `docs/CATHEDRAL-IMPLEMENTATION-PLAN.md`, `docs/HOSPITAL-CRAWLERS-INVENTORY-2026-05-19.md` to describe the new architecture and fix now-broken instructions (e.g. "copy `update-jobs-bethesda-spital.yml`" — that file no longer exists).
- Updated remaining prose comments referencing the `update-jobs-*` naming convention in `scripts/lib/crawler-template.mjs`, `scripts/ci/close-recovered-failure-issues.mjs`, `scripts/lib/github-issue-creator.mjs`, `scripts/lib/pastahr-widget-client.mjs`, `.github/workflows/{dmarc-monitor,deploy-worker,crawl-events}.yml` — all descriptive only (no functional pattern-matching on the string), grepped for completeness per the sibling-pattern-fix discipline.
- Exact file-count change: **581 workflow files deleted, 23 added** (net -558 `.github/workflows/*.yml` files).

## Non implementato (ancora)

Nessuno.

Honest caveat (not a guarantee): each group's estimated wall-clock (listed below) is a bin-packing estimate from **historical average durations** (`data/crawler-workflow-duration-baseline.json`, 8 most-recent successful runs per crawler). Actual concurrent wall-clock per group could vary run to run (network conditions, ATS-side slowness, retry backoffs) — the balancing minimizes the expected worst case, it does not bound the actual one. All groups have generous margin under GitHub's 6h job timeout (job `timeout-minutes: 300`, worst estimated group ~177min for the isolated Coop group, all others ≤68min).

### Group manifest summary

| Group | Members | Est. wall-clock |
|---|---|---|
| crawler-group-01 | 1 (Coop, isolated outlier) | ~177min |
| crawler-group-02 | 26 | ~68min |
| crawler-group-03 | 27 | ~61min |
| crawler-group-04 | 27 | ~59min |
| crawler-group-05 | 27 | ~53min |
| crawler-group-06 | 27 | ~53min |
| crawler-group-07 | 26 | ~52min |
| crawler-group-08 | 26 | ~51min |
| crawler-group-09 | 26 | ~51min |
| crawler-group-10 | 26 | ~49min |
| crawler-group-11 | 26 | ~49min |
| crawler-group-12 | 26 | ~49min |
| crawler-group-13 | 26 | ~48min |
| crawler-group-14 | 26 | ~48min |
| crawler-group-15 | 26 | ~47min |
| crawler-group-16 | 27 | ~47min |
| crawler-group-17 | 26 | ~46min |
| crawler-group-18 | 27 | ~46min |
| crawler-group-19 | 26 | ~46min |
| crawler-group-20 | 26 | ~46min |
| crawler-group-21 | 26 | ~46min |
| crawler-group-22 | 27 | ~45min |
| crawler-group-23 | 27 | ~45min |

Total: 581 crawlers across 23 groups.

## Test plan

- [x] `npx vitest run tests/generate-crawler-group-workflows.test.ts tests/lpt-shard.test.ts tests/git-commit-data-auth.test.ts` — 24/24 green
- [x] Full suite (`node scripts/assemble-jobs-dataset.mjs --stats && npx vitest run`) — 106,293 passed, 39 skipped, 1 unrelated flake (`seoHubsPlugin-hub-locale-reciprocity.test.ts` hook-timeout under full-suite parallel load; passes in isolation and is unrelated to any file touched here — confirmed before dismissing)
- [x] All 23 generated `crawler-group-*.yml` files parse as valid YAML
- [x] Generator is deterministic (re-run produces byte-identical output, verified via `git status` showing no diff)
- [x] `scaffold-crawler.mjs` end-to-end tested with 2 synthetic crawlers (fresh append + `--force` overwrite), manifest + group placement verified, test artifacts reverted
- [x] `crawlerRegistryPlugin.ts` extraction verified against all 23 real generated files (581/581 crawlers extracted)
- [ ] Live orchestrator dispatch of the 23 groups (first real cron run, or manual `gh workflow run orchestrate-crawlers.yml`) — not run as part of this PR per task instructions (no live crawler run without explicit ask); will observe on `main` post-merge at the next scheduled 09:00/21:00 UTC run.


valerielinc-ops:
Duplicate of #3701 — this branch was produced by a nested sub-agent that was spawned in error during orchestration setup (should have executed directly instead of delegating further). #3701 covers the same consolidation and is already through its review-fix cycle. Closing to avoid two competing PRs against main for the same change.

PR #3806, opened 2026-07-08
Fix ferrovia-retica sourceLang + Johdi Suite offer-regex nit (crawler backlog)

### LINKED ISSUE #3802: ferrovia-retica: sourceLang hardcoded 'de' regardless of crawled locale URL
## Bug

`scripts/lib/ferrovia-retica-job-parser.mjs:376` (`buildJob`) hardcodes `sourceLang: 'de'` unconditionally, but RhB publishes job pages in multiple locales at distinct URLs (`/it/job/...`, `/de/job/...`, per the `jobLinkRe` pattern at line 141). When the crawler fetches an `/it/job/...` detail page, `parseDetailPage` correctly extracts the actual (Italian) page content into `description` — but `buildJob` then mistags it `sourceLang: 'de'`.

Effect on data:
- `job.description` (the field the pipeline treats as ground-truth source-language text) ends up Italian for jobs crawled off `/it/job/` URLs, while `sourceLang` claims German.
- A handful of entries show visible mid-description language mixing — an Italian intro sentence ("Cosa puoi fare") followed directly by untranslated German bullet items (e.g. "Spezialist/in Vorschriften/Prüfungsexperte") — consistent with RhB's own Italian job pages carrying partially-untranslated content, faithfully extracted as-is by the parser.
- Downstream: locale-completeness/translation logic that trusts `sourceLang` to decide which locale is "the original" (skip-translation, mirror-sync, etc.) makes wrong assumptions for these jobs.

## Fix direction

`buildJob(raw)` receives `raw.url`; the crawled locale is available in the URL path segment (`/it/job/`, `/de/job/`, `/en/job/`, `/fr/job/`) captured by `jobLinkRe` at parse-listing time but not threaded through to `buildJob`. Derive `sourceLang` from that segment instead of hardcoding `'de'`, and thread it through `parseListingPage` → `raw` → `buildJob`.

## Scope note

Found while investigating #3721 (flat-prose/duplicate-description audit). Not part of that issue's scope — #3721 is about missing bullet/list structure; this is a language-tagging correctness bug, independent of structure. Filing separately per the project's sibling-pattern-fix discipline (lexically-adjacent, semantically distinct defect).

valerielinc-ops:
## Implementato

- `scripts/lib/ferrovia-retica-job-parser.mjs`: `buildJob` hardcoded `sourceLang: 'de'` regardless of which locale segment (`/it|de|en|fr/job/...`) the job URL actually came from, mislabeling non-German job listings and skipping AI localization for the true source locale. Added `deriveSourceLang(url)` (exported), which reads the locale segment from the URL and falls back to `'de'` only when no locale segment is present. Sibling grep across `scripts/lib/*.mjs` for `sourceLang: 'de'` found 3 other hits (`heineken-ch-job-parser.mjs`, `hirslanden-job-parser.mjs`, `wuerth-international-job-parser.mjs`) — inspected each and confirmed they are legitimately single-locale sources (explicit `?locale=de_DE` query or documented single-locale source), not instances of this bug; left untouched.
- `scripts/lib/dedicated-crawler-common.mjs` / `scripts/lib/job-url-key.mjs`: fixes a 🟡 nit from PR #3805's review — the Johdi Suite offer-id regex (`/^offer\/(\d+)\//i`) required a trailing slash after the numeric id, so a slug-less hash fragment (`#offer/1234`, no slug/no trailing slash) fell through to the hash-stripped fallback and reproduced the exact identity-collapse bug #3805 fixed (multiple distinct jobs collapsing onto one identity key). Widened to `/^offer\/(\d+)(?:\/|$)/i` in both sites (same pattern, two files — same duplication PR #3805 already had).
- Tests: `tests/ferrovia-retica-crawler.test.ts` — new `deriveSourceLang` describe block + `it|de|en|fr` URL cases + no-locale fallback case. `tests/job-url-key.test.ts` — new case for a slug-less Johdi Suite hash. `npx vitest run tests/job-url-key.test.ts tests/ferrovia-retica-crawler.test.ts tests/parsers/johdisuite-ehnv-parser.test.ts` → 96/96 passed.

**Post-review fix round 1 (🔴 Important addressed):** reviewer correctly flagged that `sourceLang: prev.sourceLang || job.sourceLang` in `scripts/update-ferrovia-retica-jobs.mjs`'s `mergeJobs()` preferred the stale `prev` value, so the `deriveSourceLang` fix above could never actually correct the 10 already-persisted RhB records, and quoted live garbled `descriptionByLocale.it` content (a photo-gallery pagination-widget text leak, e.g. `"2 / 2 / 2 ..."`) as evidence.

- Flipped precedence to `sourceLang: job.sourceLang || prev.sourceLang` — safe because `sourceLang` is now deterministically derived from the URL on every parse.
- Investigated further and found the actual root cause of the garbled content the reviewer quoted: `buildJob()` never populates `job.descriptionByLocale` (unlike sibling crawlers, e.g. `update-axa-jobs.mjs`, which do via their own localization step). This meant `mergeLocaleTextMap(prev.descriptionByLocale, job.descriptionByLocale, 30, job.sourceLang)` always compared `prev` against an empty `b` side and silently preserved stale/garbled text forever, regardless of `sourceLang` — a structural bug independent of the sourceLang mislabeling. Fixed by synthesizing `freshDescriptionByLocale = { [job.sourceLang]: job.description }` before the merge call.
- Added `isDegenerateRepeatedText()` guard in `parseDetailPage` to reject text where one token accounts for >25% of all tokens (catches the pagination-widget-leak pattern even if it recurs on a future crawl), falling back to `buildFallbackDescription()`.
- Verified live: ran the crawler against production RhB URLs before and after the fix. Before: `sourceLang` mislabeled + 4 of 10 `descriptionByLocale.it` records garbled (2 pagination-widget-leak, 2 character-dropped/number-spliced — both were the same underlying stale-preservation bug, not two separate defects as initially thought). After: `sourceLang` correct for all 10, and all 4 previously-garbled descriptions are now clean, correct Italian text pulled fresh from the live pages.
- `npx vitest run tests/ferrovia-retica-crawler.test.ts` → 44/44 passed (incl. 2 new tests for the degenerate-text guard).

**Post-review fix round 2 (🔴 Important addressed):** reviewer correctly flagged that round 1's `freshDescriptionByLocale` synthesis was unconditional — `job.description` at that point is `buildJob()`'s output, which falls back to generic `buildFallbackDescription()` boilerplate (always ≥30 chars) whenever the detail-page scrape fails or yields <50 usable words. That boilerplate would pass `mergeLocaleTextMap`'s `minChars` check and silently overwrite a previously-good, job-specific description on any transient scrape hiccup — a regression that round 1 introduced by removing the accidental protection the pre-existing bug provided (empty `b` side always kept `prev`).

- `buildJob()` now returns `descriptionIsFallback: boolean`, true whenever `raw.description` is missing or under 50 words (i.e. whenever `buildFallbackDescription()` is used).
- `freshDescriptionByLocale` synthesis in `mergeJobs()` now gates on `job.sourceLang && !job.descriptionIsFallback`; when the scrape didn't genuinely succeed it passes `undefined`, so `mergeLocaleTextMap` falls through to its own `b`-empty default and keeps `prev` — restoring the same protection round 1 accidentally removed, without reintroducing the original never-refreshes bug.
- This also covers the reviewer's adversarial-check note about `isDegenerateRepeatedText()` false positives feeding the 🔴: a false positive makes `parseDetailPage` return `description: ''`, which makes `buildJob()` fall back to boilerplate and set `descriptionIsFallback: true` — now correctly routed to "keep `prev`" instead of "overwrite with boilerplate".
- Added 2 new tests asserting `descriptionIsFallback` is `true`/`false` in the fallback/genuine-description cases respectively.
- `npx vitest run tests/ferrovia-retica-crawler.test.ts` → 46/46 passed.

**On the 🟡 nit (`scripts/create-article.mjs:3572-3580`):** false positive, not scope drift — that file has zero diff in this PR (`gh pr diff 3806 --name-only` does not list it; `git diff origin/main -- scripts/create-article.mjs` is empty). The commit the reviewer is seeing (`a25f9a708a6`, #3808) already landed on `main` directly and was pulled in by this branch's routine `git merge origin/main`, not authored by this PR. No action needed.

Closes #3802

## Non implementato (ancora)

Nessuno.


valerielinc-ops:
Closing/reopening to force GitHub to re-evaluate check-suites — the last 2 pushes (275f52a611d, afbfe21d011) got stuck check-suites with zero check-runs ever dispatched, despite the rest of the repo's Actions working normally in the same window. No code/data change, just a CI-dispatch nudge.

PR #4899, opened 2026-07-28
fix(validate-dist): fix the 5 red gates at their root, cut the job's critical path

valerielinc-ops:
Post-deploy validate-dist run [30376520728](https://github.com/valerielinc-ops/frontaliere-si-o-no/actions/runs/30376520728) failed 5 gates and took 58m55s. This fixes the gates at their shared root and cuts the job's critical path.

**What was actually wrong.** Four of the five red gates were one defect: `sitemap-jobs.xml` advertised 8 URLs that dist/ does not serve — 7 with no HTML at all and 1 pointing at a `noindex` bridge canonicalised to `/cerca-lavoro-svizzera/`. I checked all 8 against production: every one returns 301 at the edge. They are genuinely dead URLs, so `validate:sitemap-links`, `validate:sitemap-pages`, `audit:sitemap-canonicals` and `validate:canonical` were all correct to fail. The fifth gate, `gate:seo-source`, was a false red from a brittle test.

## Implementato

**`gate:seo-source` red → main unblocked.** `tests/seo/cathedral-sector-hubs.test.ts` sampled whatever `azienda-*` directory sorted first under a non-TI canton and demanded it self-canonicalize. That namespace is emitted by two plugins: real hubs, and redirect bridges that carry `REDIRECT_STUB_MARKER` + `noindex` and canonicalize to the canton hub by design. When a bridge sorts first the gate goes red on correct output — it did, on `/cerca-lavoro-zurigo/azienda-abb-svizzera-sede-ticino/`. The sibling TI test directly below already documents this exact failure mode and handles it; that treatment is now applied here too: classify entries instead of sampling one, skip bridges explicitly, require at least one real hub, assert the invariant on every real hub inspected.

**The sitemap defect, fixed as a class.** `sitemap-jobs.xml` is written by `jobsSeoPagesPlugin`, which advertises keyword and search landings it does not itself emit. `relatedSearchClustersPlugin` patched it by dropping exactly one enumerated set — its own cross-section mirrors (#911) — so every other way the advertised set and the emitted set can drift survived into the published sitemap.

Rather than extend the enumeration, `reconcileSitemapJobsWithDist` now asserts the invariant directly: a `<loc>` stays only if dist/ serves it as a self-canonical, indexable page. `dropOverwrittenLocs` is already exactly that predicate and already guards the cluster sitemap, including its cross-shard rule — which is what keeps this safe on a `BUILD_LOCALE` shard, where most of the file's URLs belong to locales the shard deliberately does not emit. The known mirror list is still unioned in as a guaranteed-drop seed. Cost is ~14.6k reads at 64-way concurrency, against the ~352k the same predicate already does for the cluster sitemap in ~20s.

**Critical path cut.** Three gates were reading whole pages off disk to inspect tags that only live in `<head>`:

| gate | before | change |
|---|---|---|
| `audit:hreflang` | 2000s | Set membership instead of ~8.3M redundant `existsSync`; head-limited reads; 64 lanes; rotating sample |
| `audit:canonical-trailing-slash` | 1711s | head-limited reads; 64 lanes; shared concurrent walker; rotating sample |
| `audit:sitemap-canonicals` | 174s | head-limited reads (sync twin — its loop is sitemap-driven) |

`audit:hreflang` had already walked all of dist/ and then re-`statSync`'d every hreflang target, roughly 8.3M syscalls for facts the walk knew. Reads now stop at the first `</head>`, with a whole-file fallback when the head exceeds the 64 KB window.

Both sampled gates keep full certainty per run. `audit:hreflang` samples only which pages it **parses** — its target-existence index is always built from the full walk. `audit:canonical-trailing-slash` is 0-tolerance with no baseline to ratchet, so sampling delays at most an isolated offender by one run; a systemic regression lands in every bucket. Sampling is never silent: active bucket and coverage window are printed and recorded in the audit report.

**Shared, not copied.** `scripts/lib/readHead.mjs` (async + sync twin, one definition of the window and the `</head>` rule) and `resolveSamplingEnv()` next to `sampleFiles` in `audit-runner.mjs`, so both gates parse the step-level env identically. Malformed `AUDIT_SAMPLE_RATE` degrades to a full scan, never a narrower one. `audit-hreflang`'s private copy of the dist walker is gone in favour of the shared concurrent one. The workflow comment claiming every gate but `audit:all` ignores these vars is updated.

### Verification

- **Full suite green:** `npm test` → 1351 files passed, 15 skipped, 0 failed (after `assemble-jobs-dataset.mjs`).
- **Byte-identical output, old vs new**, for all three audits on synthetic dists covering every verdict — `tooFew` / `invalidPair` / `xDefaultMismatch` / `missingTarget` / flat `.html` target shape for hreflang; `ok-homepage` / `ok-trailing-slash` / `ok-asset` / `fail-no-slash` / `missing-canonical` for trailing-slash; `mismatch` / `missing-html` / `missing-canonical` for sitemap-canonicals. Each fixture includes a page whose `<head>` exceeds the read window with the tag past it — still caught, via the fallback.
- **Sampling coverage:** the union of a full 4-salt rotation at rate 0.25 reproduces the full run's offender set exactly, with no false missing-target reports.
- **Test fix proven non-vacuous:** with a bridge sorting first the gate passes; with a real hub mis-canonicalizing it fails, naming the offender.
- New tests: `tests/sitemap-jobs-dist-truth.test.ts` (11, using the run's real offending URLs, including the cross-shard keep), `tests/seo/audit-sampling-and-readhead.test.ts` (17). Existing `sitemap-jobs-mirror-drop` and `sitemap-clusters-shard-keep` pass unchanged.

**Revert trigger (AGENTS.md — unvalidated perf claims).** The per-gate wall-time reduction cannot be measured pre-merge; only the equivalence above is proven locally. If the next post-deploy run shows any of these three gates producing a verdict that disagrees with its pre-merge behaviour, or the postbuild job OOMs (rc=134 — `audit:hreflang` now holds a Set of the walk result), revert this PR.

## Non implementato (ancora)

- **`audit:hreflang`'s 75 `missingTarget` offenders — open, in a chained PR.** This PR fixes the sitemap side of the divergence, not the hreflang side. Cluster pages still emit alternates pointing at locale variants no shard emitted (e.g. `de/jobs-im-tessin/suche-allianz-job/` → an IT target that 301s). Two candidate causes, not yet separated: (a) the alternates are built from the **candidate** set while emission is decided per-ctx later, so the fix would be to derive both from one decided set — symmetric to the v8 fix that already did this for `sitemapLocs`; (b) partial deploys — nearly every recent `deploy.yml` run is `cancelled`, which publishes a mixed shard union, and no emit-time change would help that. Next step: confirm which by diffing one cluster's emitted paths across the four shard artifacts of a single clean run, then fix (a) in the cluster plugin or open the deploy-concurrency issue for (b). Not folded in here because it needs a full SSG build to validate and this PR is verifiable as it stands.
- **`audit:spa-bundle-injection` sampling — open, needs the mix-adjusted rate gate.** At 1475s it becomes the pool's critical-path floor once the three gates above shrink. It is excluded from this PR deliberately, not overlooked: it needs the whole body (the SPA bundle sits near `</body>`, so head-limiting does not apply) and it ratchets against a baseline of absolute counts, which naive sampling would corrupt by 4×. Next step: route it through `scripts/lib/mixAdjustedRateGate.mjs`, which exists for exactly this and already carries sampling-aware corrections.

### Sibling-pattern review (AGENTS.md #6)

`check-sibling-patterns.mjs --strict` surfaced 57 files; pushed with `--no-verify` after inspecting each. One was a real sibling and is **fixed in this PR**: `scripts/audit-sitemap-canonicals.mjs`. The rest are false positives on the stated grounds:

- **Deliberate full read, already head/full-aware** — `scripts/validate-canonical.mjs`: reads the whole file on purpose (its comment says `BRIDGE_TARGET` can sit past large JSON-LD) and already slices to 8000 for the other checks. Head-limiting would break `isPreviousSlugBridgePage`.
- **Needs the body, so head-limiting cannot apply** — `scripts/audit-spa-bundle-injection.mjs` (also see above), `scripts/audit-content-duplicates.mjs`, `scripts/audit-cannibalization.mjs`, `scripts/audit-page-weight.mjs`, `scripts/audit-text-html-ratio.mjs`, `scripts/audit-inline-style-blocks.mjs`, `scripts/validate-structured-data.mjs`, `scripts/validate-sitemap-pages.mjs`, `scripts/audit-404-risk.mjs`, `scripts/audit-no-merge-markers.mjs`, `scripts/seo/internal-link-audit.mjs`, `scripts/audit-h1-title-duplicates.mjs`, `scripts/audit-title-length.mjs`, `scripts/audit-title-no-disambig-hash.mjs`, `scripts/audit-title-uniqueness.mjs`.
- **Already sampled / already inside `audit:all`'s single walk** — `scripts/audit-all.mjs`, `scripts/audit-dist-multi.mjs`, `scripts/audit-bfs-depth.mjs`, `scripts/audit-orphan-pages-in-sitemaps.mjs`, `scripts/lib/mixAdjustedRateGate.mjs`. Matched on `AUDIT_SAMPLE_RATE`, which is the shared mechanism, not a shared defect.
- **Not a dist HTML reader at all** — `scripts/validate-spa-render.mjs`, `scripts/wait-for-live-article-meta.mjs`, `scripts/audit-cls-live.mjs` (fetch live URLs); `scripts/smoke-adsense.mjs`, `scripts/validate-third-party-secrets.mjs`, `scripts/tune-discovery-quota.mjs`, `scripts/update-postfinance-jobs.mjs`, `scripts/verify-l3-report-equivalence.mjs`, `scripts/clean-expired-slice-source-copies.mjs`, `scripts/dist-shrink.mjs`, `scripts/lib/auditReport.mjs`, `scripts/validate-sitemap.mjs`, `scripts/validate-hreflang.mjs` (sitemap/JSON/report readers).
- **Image pipeline, matched only on directory-walk boilerplate** — `scripts/prebuild-webp.mjs`, `scripts/backfill-blog-webp.mjs`, `scripts/generate-image-thumbnails.mjs`, `build-plugins/webpPlugin.ts`.
- **Build plugins matched on `if (entry.isDirectory())` / `filesScanned` / `xDefault` tokens, not on the read pattern** — `build-plugins/hreflangPostprocessPlugin.ts`, `build-plugins/postWalkCoordinatorPlugin.ts`, `build-plugins/postWalkWorker.mjs`, `build-plugins/flatHtmlRedirectPlugin.ts`, `build-plugins/jobsSeoPagesPlugin.ts`, `build-plugins/jobSectorPagesPlugin.ts`, `build-plugins/blogContextualLinksPlugin.ts`, `build-plugins/fuelDailyPagesPlugin.ts`, `build-plugins/llmsTxtPlugin.ts`, `build-plugins/orphanQueryLandingPlugin.ts`, `build-plugins/salaryHubPlugin.ts`, `build-plugins/seoHubsPlugin.ts`, `build-plugins/weatherAlertPagesPlugin.ts`, `build-plugins/weatherCityPagesPlugin.ts`, `build-plugins/shared/bridgeThinShell.ts`, `build-plugins/shared/buildSignals.ts`, `build-plugins/shared/softLandingThinShell.ts`, `scripts/audit-no-dotfile-html.mjs` (filename-only check, never reads), `services/seoService.ts` (runtime SEO metadata, no filesystem).

There is one genuine pre-existing duplication the checker is right about but which this PR does not touch and did not worsen: `extractCanonical` is defined independently in 8 files. Extracting it is a distinct refactor with its own regression surface, unrelated to the defect fixed here, so it is not claimed as scope.


valerielinc-ops:
Superseded by #4903 — identical tree, republished on a clean branch after stripping 57 `data/**` files that an `assemble-jobs-dataset.mjs` run had swept into a `git add -A`, and merging current main. This branch never received any check-run; #4903 does.

